PR #9, opened 2013-10-08
DATACOUCH-36 - Don't allow NULL values to be stored.

dharrigan:
It is illegal to store a NULL value as a document. An attempt to do so will
throw an IllegalArgumentException.

-=david=-


src/main/java/org/springframework/data/couchbase/core/mapping/CouchbaseDocument.java
```diff
@@ -259,15 +259,19 @@ public CouchbaseDocument setId(String id) {
...
-   * @param clazz the class type to check and verify.
+   * Objects that are NULL cannot be stored.
+   *
+   * @param value the object to verify its type.
    */
-  private void verifyValueType(final Class<?> clazz) {
+  private void verifyValueType(final Object value) {
+    if(value == null) {
+      throw new IllegalArgumentException("Attribute of type null cannot be stored.");
+    }
```
prettyvoid: Any idea why I can't persist null fields (not null documents) using a CrudRepository? I want to save null fields.

PR #24, opened 2013-12-30
Configuration for updating expire settings for documents when read is ma...

awislowski:
Configuration for updating expire settings for documents when read is made. Configuration is based on Document annotation. When read of document takes place then touch action is executed on document. It is an asynchronous action so it does not block read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ This project adds common Spring Data functionality (like configuration, template
 Full documentation is still in the making, so this README outlines the basic steps you need to do in order to get it up and running. It also provides a very rough overview of its capabilities. Note that this library is still in the making and does not provide all features yet.
 
 ## Features
-### Implemented
+### Implemented 1.0
 
  - Templates
  - JavaConfig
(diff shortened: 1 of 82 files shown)


simonbasle:
hi @awislowski! sorry this fell through the cracks :skull: :scream:
I looked at what seemed the last commit actually related to this (6f5ce5c) and the discussing in jira ticket and this change has good potential... **I'd like to merge it, but the PR is outdated and I'd prefer to implement this in `2.0.x`.**

Would you be willing to try a new PR to port it in the 2.0.x branch? :recycle: :pray: (I feel the size of the actual change is small enough that this is doable with copy pasting and small edits). Otherwise I can do it myself.

Caveats:
 - it's still missing a bit of javadoc
 - the `findByView` would need a different approach in 2.0: currently it doesn't call `bucket.get` but uses the row's `document(RawJsonDocument.class)` method.
 - there is a new `findByN1ql` method to cover as well... which is trickier because it doesn't even call KV operation, just uses the already SELECTed content from the N1QL result... So here we'd probably need an async touch in a fire-and-forget fashion. Or defer work on that to another ticket, but more and more custom repository methods would rely on N1QL and thus omit the touch...

awislowski:
Hi @simonbasle,

I have not much time for work concerning couchbase, as I am not using it at work nowadays.

I have created new PR, as you proposed: https://github.com/spring-projects/spring-data-couchbase/pull/96

and I am closing this one.

PR #32, opened 2014-09-24
DATACOUCH-110 - Use CustomConversions for Map/Collection values

winard:
This patch fix the mapping issue in DATACOUCH-110.


daschl:
HI @winard I just merged another PR regarding DATACOUCH-110.. I know this has been some time out, but if it still concerns you - you can check it on master! It will be part of fowler, so I'm closing this out... your tests have been also merged in, thanks for the contribution!

PR #37, opened 2015-04-02
fixed potential typo in readme.md

KlausUnger:
potentially mixup of attributes - the view backing the findByLastname method should emit lastname not firstname


simonbasle:
:+1: that looks more correct

PR #38, opened 2015-04-15
Fixing Readme URL's

kdheerendra:
URL's moved so accordingly changing the url for user guide and java docs.


simonbasle:
:+1: thanks

PR #40, opened 2015-06-16
prepare bare template class, use preview of SDK 2.2.0

simonbasle:
template will be in same package as 1.x generation, that is "core"


daschl:
+1

PR #41, opened 2015-06-16
basic configuration (xml parsers and @Configuration)

simonbasle:
For xml configuration, only the basic parsers have been implemented.
Prepared an `AbstractCouchbaseConfiguration` `@Configuration` that allows to create beans for the `CouchbaseEnvironment` (including sharing one from outside of the Spring context), the `Cluster` and the `Bucket`.
Skeletal implementations of `CouchbaseOperations` and `CouchbaseTemplate` for compilation purposes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
 
     <dist.key>DATACOUCH</dist.key>
 
-    <couchbase>2.1.3</couchbase>
+    <couchbase>2.2.0-dp</couchbase>
     <jackson>2.3.2</jackson>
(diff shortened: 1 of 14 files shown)


simonbasle:
@daschl can you :+1: this?

daschl:
@simonbasle for now, can you make the config part simpler so that it is just managed? We'll figure out the proper strategy from there. otherwise I'm good

simonbasle:
I simplified the overrides needed for environment management + changed the `translationService` bean name

daschl:
+2

PR #42, opened 2015-06-22
[2.0] xml configuration for env, cluster, bucket

simonbasle:
This PR implements parsers (and FactoryBeans as needed) for the environment, cluster and bucket definitions.
Most tunings on the environment are possible through the xml configuration, except a few that are both more complicated to implement and obscure enough not to warrant the possibility (observeIntervalDelay, reconnectDelay, retryDelay, userAgent, packageNameAndVersion, ioPool, scheduler and eventBus).


daschl:
:+1:

PR #48, opened 2015-06-29
[SDK2] fix template constructor order + improve tests

simonbasle:
integration tests have been better separated from unit tests (and each have their own configuration).
new integration tests have been put in place for templateParser and monitor classes.


simonbasle:
@daschl please review when you can :bowtie:

daschl:
:+1:

PR #52, opened 2015-07-13
DATACOUCH-136 - Add @N1QL query annotation.

simonbasle:
The annotation allows to execute N1QL queries that are provided inline. For ease of use, one can use the placeholder $SELECT_ENTITY$ instead of writing the SELECT and FROM clauses.

QueryParams passed in as a method parameter will be used to enrich the Query. Other parameters on the annotated method will be used as values for positional placeholders in the statement.

The template's findByN1QL now expect enough data to reconstruct the full CouchbaseDocument (including ID and CAS), the ad hoc version of it has been renamed findByN1QLProjection.

### DIFF
--- a/src/integration/java/org/springframework/data/couchbase/repository/UserRepository.java
+++ b/src/integration/java/org/springframework/data/couchbase/repository/UserRepository.java
@@ -18,6 +18,7 @@ package org.springframework.data.couchbase.repository;
 
 import com.couchbase.client.java.view.ViewQuery;
 
+import org.springframework.data.couchbase.core.view.N1QL;
 import org.springframework.data.couchbase.core.view.View;
 
 /**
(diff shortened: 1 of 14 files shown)


simonbasle:
@daschl please code review :)

src/integration/java/org/springframework/data/couchbase/repository/UserRepository.java
```diff
@@ -28,4 +29,10 @@
   @View(designDocument = "user", viewName = "all")
   Iterable<User> customViewQuery(ViewQuery query);
 
+  @N1QL("$SELECT_ENTITY$ WHERE username = $1")
```
daschl: is this just a manual test? I guess this should be auto generated by us, right? the @N1QL would just be to override the default, or..?

daschl:
+1

PR #53, opened 2015-07-13
DATACOUCH-137 - queryDerivation for N1QL annotated methods.

simonbasle:
Abstracted the N1QL-based variants in AbstractN1qlBasedQuery. This implementation doesn't recognize QueryParams or QueryPlan anymore in the method parameters so that the method signatures are not store-dependent.

Added query derivation from PartTree to N1QL query, using N1qlQueryCreator.

Unit tested the mapping between a Part.Type and the corresponding N1QL Expression.


simonbasle:
@daschl please code review after #52 (should have only DATACOUCH-137 commit in there)

daschl:
+1, get that in

PR #56, opened 2015-07-13
DATACOUCH-141 - Change query lookup order and make N1QL default.

simonbasle:
The @N1QL annotation was changed to @Query, for discoverability and also
convey that this is now the preferred method.

Order is now view if @View present, N1ql inline if @Query present with
a Statement, N1ql query derivation if @Query without attribute or no
annotation.

@View can be without attribute in order to compute the view
and design document from method name.


simonbasle:
@daschl please review :)

daschl:
+1

PR #58, opened 2015-07-16
DATACOUCH-140 - Use Converter to match query params and stored values.

simonbasle:
When deriving a query, use the CouchbaseConverter to transform the parameters for fields that would be converted upon storage.

Reworked the DateConverters so that reading conversion is done from any Number (since N1QL may return longs in Double scientific notation for example).


simonbasle:
@daschl please code review :sunglasses:

daschl:
one note on the converter consistency naming on decoding, other than looks good.

PR #60, opened 2015-07-20
DATACOUCH-144 - Detect N1QL dependency and fail fast if not available.

simonbasle:
Detect N1QL is not available at configuration time for the repositories and at runtime for the template. This throws a UnsupportedCouchbaseFeatureException.

The ClusterInfo is captured at bootstrap and used to determine if the feature is available. The xml configuration will need the ClusterInfo-related credentials to be provided explicitly.


simonbasle:
pinging @daschl for code review :eyeglasses:

daschl:
+1

PR #61, opened 2015-07-20
DATACOUCH-145 - Use Id/Field annotations from Couchbase SDK

simonbasle:
The @Field annotation has been removed in favor of the one from the Couchbase SDK.

Similarly, CouchbasePersistentEntity now recognize the SDK's @Id annotation. Note that if both Spring Data and Couchbase @Id are present (on two distinct fields), the one from Spring Data will be taken into account.


daschl:
:+1:

PR #64, opened 2015-07-27
DATACOUCH-148 - Allow to configure repository queries staleness globally

simonbasle:
The staleness to apply on generated view queries (`Stale`) and N1QL queries (`ScanConsistency`) can now be chosen via the configuration.

It is accessed from the `CouchbaseOperations` interface but is applied to the template only. In xml, the `viewStaleness` and `n1qlStaleness` attributes are configurable on the `couchbase:template` element (string value of the enums to be passed in).


simonbasle:
@daschl please code review :book:, @olivergierke I'd like your take on that as well (maybe there's a better "Spring Way ©" to do it :panda_face:)

daschl:
@simonbasle would it be possible to converge on a sdc-based enum for consistency which in turn translates over into view and n1ql staleness params? I know we dont have that in the SDK, but I was thinking for a user perspective it might help them

daschl:
Like "Consistency.STRONG or EVENTUAL" or something

simonbasle:
@daschl added a `Consistency` enum that does just that. Also edited the documentation

simonbasle:
squashed in preparation for cherry-picking into 2.0.x

PR #65, opened 2015-07-30
DATACOUCH-154 - Guess design doc as uncapitalized class name.

simonbasle:
In the SimpleCouchbaseRepository the guessed design document was lowercase of the entity class, whereas in ViewBasedCouchbaseQuery it is uncapitalized (only first letter is lowercase).

This has been made consistent, using the uncapitalize method. The doc now reflects that and explicitly have examples with camel case classes (UserInfo instead of User).


simonbasle:
@daschl please review :)

daschl:
:+1:

PR #66, opened 2015-07-31
DATACOUCH-156 - Make generated N1QL queries filter on the type field.

simonbasle:
The generated N1QL queries currently don't filter on the type at all.

Add a criteria to the WHERE clause that checks the field holding type information is matching the entity fully qualified class name.

Add a placeholder for inline N1QL queries that can be replaced by the same type information criteria.

### DIFF
--- a/src/integration/java/org/springframework/data/couchbase/repository/Item.java
+++ b/src/integration/java/org/springframework/data/couchbase/repository/Item.java
@@ -1,0 +1,38 @@
+package org.springframework.data.couchbase.repository;
+
+import com.couchbase.client.java.repository.annotation.Field;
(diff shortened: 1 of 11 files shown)


simonbasle:
@daschl can you review this (quite important one!). I built this upon another commit, 6019609, that I pushed directly.

daschl:
:+1:  does that impact docs?

simonbasle:
ah you're right, it's visible to the user in the inline statement mode, which is very quickly documented, amending...

simonbasle:
@daschl updated doc

daschl:
:+1:

PR #68, opened 2015-09-09
DATACOUCH-162: CouchbaseTemplate should not cast values retrieved from couchbase to String

Aloren:
https://jira.spring.io/browse/DATACOUCH-162

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
@@ -128,7 +128,7 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationEventP
   }
 
 
-  private CouchbaseStorable translateDecode(final String source, final CouchbaseStorable target) {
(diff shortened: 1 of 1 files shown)


simonbasle:
Hi @Aloren, thanks for that PR!

Can you make the use case (using a custom `TranslationService` that doesn't operate with a JSON `String`) more explicit by putting in a comment and a unit test with such a custom service?

`JacksonTranslationService` will cast the entry to `String`, so that should maybe be made more explicit as well in the javadoc, what do you think?

See also the JIRA issue for more comments on the "why" vs the "how"

Aloren:
@simonbasle Hi!

Ok, I'll do test with binary data in 1.3.x version.

By the way maybe it makes sense to do support for such thing in 2.0 version as well? As you said in jira issue, couchbase relies on json format, but still it is possible to save data in binary format (of course you will not be able to use couchbase views etc). For big objects serialization/deserialization may be an issue and migrating to something like cryo is a great option. So what do you think? Thanks.

Aloren:
Hi! I suppose current pull request can be closed. Thanks.

PR #71, opened 2015-09-21
DATACOUCH-164 - Use stable Couchbase SDK

simonbasle:
Make modifications due to name changing in 2.2.0, the latest stable release of the Couchbase SDK.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
 
         <dist.key>DATACOUCH</dist.key>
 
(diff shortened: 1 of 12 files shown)


simonbasle:
@daschl anything else in the latest release that you think could impact SDC?

daschl:
@simonbasle don't think so - we had mainly those renames, so if it compiles we should be good for now.

simonbasle:
compiles and passes tests, so I'll merge ;)

PR #72, opened 2015-09-25
DATACOUCH-165 - Add Geo Support

simonbasle:


simonbasle:
pinging @daschl, I think this is ready for prime time

simonbasle:
note: I will improve on this in another PR where I may rely on tools like `JTS` for polygon intersections and removing false positives (and also compute bounding boxes)

simonbasle:
didn't find a way to easily plug false-positive elimination in the framework (since it has to be applied on the correct property after the results have been served by the spatial view, and at this point this kind of information is lost).

offering them as utility class `AwtPointInShapeEvaluator` and `JtsPointInShapeEvaluator`.

open to suggestions on naming and if you have better ideas to do that

simonbasle:
will squash into two commits: one for base geo implem and second for false positives utility class

PR #73, opened 2015-10-02
DATACOUCH-161 - Support repos with different buckets

simonbasle:
Added a configuration bean that allows to express what `CouchbaseOperations` each repository should use, individually or by domain type.


simonbasle:
pinging @daschl for code review :computer: :eyes:

simonbasle:
also @christophstrobl and/or @olivergierke I'd like your quick review too.
this is the best way I found how to do it (the `@EnableCouchbaseRepositories` cannot be repeated and there's no practical ways of declaring such a mapping on it that I could think of, since annotations are limited a lot, String/enums/constant values/... and single depth arrays of these)

daschl:
@simonbasle looks good to me, but I'd prefer if @christophstrobl and/or @olivergierke  can sign this too, so we are doing it the spring way ;)

daschl:
:+1:

simonbasle:
squashed and rebased commits, now merging

PR #74, opened 2015-10-14
DATACOUCH-158 - Allow more explicit expiry unit

simonbasle:
The Document annotation now also support explicitly giving a TimeUnit for the expiry, ensuring the framework will correctly convert that to the expected Couchbase format.

pinging @daschl for review :smile:


daschl:
:+1:

PR #75, opened 2015-10-19
DATACOUCH-166 - Add basic SpEL support

simonbasle:
first pass at having SpEL support in inline N1QL queries


simonbasle:
pinging @daschl for second pair of eyes

simonbasle:
@daschl I've updated the PR to use SpEL for N1QL placeholders, and improved documentation.

PR #76, opened 2015-10-20
DATACOUCH-168 - Support and document all reduce functions

simonbasle:
Detection whether or not reduce should be activated is now centralized. Triggers are:
  - using the `count` prefix instead of `find` in method name
  - setting `reduce = true` in the `View` annotation explicitly.

The reduce can thus be returning any kind of value (added an integration example with `_stats`), which is now properly documented.


daschl:
:+1: looks good to me.

PR #77, opened 2015-10-23
DATACOUCH-169 - Automatic index management

simonbasle:
Main indexes (ie indexes that can be used by repositories to retrieve ALL the documents of a particular entity type) are covered.

Repository interfaces can be annotated with an annotation for each type of main index that can be automatically created.

The IndexManager is responsible for creating said indexes (potentially skipping creation if an index is already in place).

Added doc and integration test.

Renamed package core.view to core.query since it now contains various non-view related annotations and classes.


simonbasle:
ready for review @daschl

simonbasle:
@daschl made some modifications re your comments (N1qlSecondaryIndexed should not deal with PRIMARY at all, it should be possible to toggle off the feature depending on the context at runtime)

simonbasle:
just rebased over #80

simonbasle:
@daschl ready for review :book:

daschl:
reviewed the latest xml changes, :+1:

simonbasle:
rebased and squashed to be merged in `2.0.x`

PR #78, opened 2015-11-13
DATACOUCH-172 - Fix PagingAndSorting consistency

simonbasle:
Default consistency from template wasn't used in PagingAndSorting findAll methods.


daschl:
:+1:

PR #79, opened 2015-11-13
DATACOUCH-161 - Improve multitemplate mapping configuration

simonbasle:
The configuration step made it look like a new mapping was necessary in overrides, whereas it is actually possible and clearer to mutate the base mapping.

Added a convenience method to get the MappingContext from the repository mapping.


daschl:
looks good, see nothing wrong with it that way :) makes it clearer for the user :+1:

PR #83, opened 2015-11-17
DATACOUCH-176 - Geo false positive autoelimination

simonbasle:
So far with geo queries a bounding box approximation was performed on Circle and Polygon queries (since Couchbase internally only supports Bounding Box queries).

It was confusing since in previous iteration it was up to the user to remove these false positives "after the fact" in code calling the repository.

This change allows for the framework to retain enough information to perform false positives elimination automatically and transparently for the user.


daschl:
:+1:  looks good to me given, that as @simonbasle assured me all the awt stuff is available in every jvm :)

PR #86, opened 2016-01-08
DATACOUCH-186 - Index-backed methods safe from deleted stales

simonbasle:
This commit ensures that methods in the template and the repository base classes are correctly ignoring documents that have been deleted but are still identified by the view due to using weak consistency (Stale.TRUE).

FindByView in template now ignores null rows in its result list. It also correctly forces the includeDocs parameter in order not to risk a bad transcoding of the document (javadoc clarified).

DeleteAll in the repository CRUD implementation will correctly detect attempts at deleting a non-existing document (DocumentDoesNotExistException is also correctly mapped to a Spring Data exception now).

DeleteAll behavior with DataRetrievalFailureException is tested in RepositoryIndexUsageTest.

### DIFF
--- a/src/main/asciidoc/repository.adoc
+++ b/src/main/asciidoc/repository.adoc
@@ -379,6 +379,7 @@ One aspect that is often needed and doesn't have a direct equivalent in the Spri
 data, because the latest version hasn't been indexed yet. This gives the best performance at the expense of consistency.
 
 Note that weaker consistencies can lead to data being returned that doesn't match the criteria of a derived query.
+One trickier case is when documents are deleted from Couchbase but views have not yet caught up to the deletion. With weak consistency this can mean that a view would return IDs that are not in the database anymore, leading to null entities. The `CouchbaseTemplate`s `findByView` and `findBySpatialView` methods will remove such stale deleted entities from their result in order to avoid having nulls in the returned collections. Similarly, `CouchbaseRepository`'s `deleteAll` method will ignore documents that the backing view provided but the SDK remove operation couldn't find.
 
 If one wants to have stronger consistency, there are two possibilities described in the next sections.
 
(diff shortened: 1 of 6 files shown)


simonbasle:
pingin @daschl for review

src/main/asciidoc/repository.adoc
```diff
@@ -379,6 +379,7 @@ One aspect that is often needed and doesn't have a direct equivalent in the Spri
 data, because the latest version hasn't been indexed yet. This gives the best performance at the expense of consistency.
 
 Note that weaker consistencies can lead to data being returned that doesn't match the criteria of a derived query.
+One case that is even trickier is when documents are deleted from Couchbase but views have not yet caught up to the deletion. With weak consistency this can mean that a view would return IDs that are not in the database anymore, leading to null entities. The `CouchbaseTemplate`s `findByView` and `findBySpatialView` methods will remove such stale deleted entities from their result in order to avoid having nulls in the returned collections. Similarly, `CouchbaseRepository`'s `deleteAll` method will ignore documents that the backing view provided but the SDK remove operation couldn't find.
```
daschl: - "is even tricker is" sounds funny, maybe you can reword it a bit?

src/main/java/org/springframework/data/couchbase/repository/support/SimpleCouchbaseRepository.java
```diff
@@ -180,7 +182,12 @@ public void deleteAll() {
 
     ViewResult response = couchbaseOperations.queryView(query);
     for (ViewRow row : response) {
-      couchbaseOperations.remove(row.id());
+      try {
+        couchbaseOperations.remove(row.id());
+      } catch (DataRetrievalFailureException e) {
+        //ignore stale deletions
+        if (!(e.getCause() instanceof DocumentDoesNotExistException)) throw e;
```
daschl: isn't that now mapped to the Data retreival exception?

simonbasle: yes but that's the one caught just above ;)
so basically I check the cause of this `DataRetrievalFailureException` to indeed be a `DocumentDoesNotExistException` (which is a bit defensive for now, but it could be another cause)

daschl: :+1:

PR #88, opened 2016-01-15
DATACOUCH-189 - Test and document how to change repo implementation

simonbasle:
Added an integration test for custom method customization.
Moved the integration test for base class customization.
Put the samples in the documentation in a new "Changing Repository Behaviour" section.


daschl:
just a minor note, otherwise :+1:

PR #90, opened 2016-01-25
DATACOUCH-194 add ablity for sorting and database backed pagination

chudson-ixxus:


simonbasle:
Hi @chudson-ixxus, and thanks for the PR.

I cannot even look at this change right now, the diff makes it look like the whole file changed... Could you please rework it so the formatting difference is minimal and only on modified lines, and we'll take it further from there?

I don't have a formatting rule configuration to export but please deactivate your formatting rule on the whole file (you should be able to at least only activate it on modified lines). Javadocs have hard-wrap breaks for readability in the code for instance...

Thanks!

chudson-ixxus:
yes sorry about that, was in a rush and only noticed the autoformatting issue when I created the PR. I have now removed the formatting.
let me know if you want anything else changing

simonbasle:
still a little bit rough to read :) instead of a revert commit, could you squash the two? Here is a quick way of doing it using interactive rebase if you need a reference:

```bash
git rebase -i HEAD~2
```
An editor will open with 1 line per commit you are rebasing (here 2). Locate the line for commit c4f56a6 (revert autoformating) and replace the instruction `pick` at the beginning with `fixup`. Save and close the editor. At this point, your branch history has been rewritten with the two commits merged into 1 commit. Quickly verify the code looks in good shape, then use the following command to update this PR:

```bash
git push --force-with-lease
```

(`--force-with-lease` is a slightly less destructive option than `--force` as it will abort if another incompatible commit has been pushed since your last pull)

simonbasle:
My main concern with this kind of change is that the inline query could **already** contain an ORDER BY , LIMIT or OFFSET clause... However you cannot just guard against that by checking `contains("ORDER BY")` for instance, because these words could be escaped inside the query...

Eg. "`SELECT * FROM bucket WHERE type = "queries" AND content LIKE "%ORDER BY%"`" is eligible for using a `Pageable`
While "`SELECT * FROM bucket WHERE type = "queries" ORDER BY lastExecuted DESC`" is not.

pivotal-issuemaster:
@chudson-ixxus Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=90) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

subalakr:
Fixed by PR #128. Closing.

PR #91, opened 2016-01-27
DATACOUCH-187 - Allow counting queries and single projections.

simonbasle:
This change allows simple projections to types long, boolean, String when an inline N1QL query uses an aggregation function like AVG, COUNT.

The projection's result must be a single row with a single key/value pair in the N1QL response. Only the value gets returned.


simonbasle:
mmh I merged this one but didn't get a review from @daschl, can you have a look and tell me if you find anything fishy? :bow:

PR #92, opened 2016-01-27
DATACOUCH-192 - Prefix all bean names with "couchbase"

simonbasle:


simonbasle:
pinging @daschl for review

daschl:
:+1: , but I'm curious: can you tell me what the motivation for that is?

simonbasle:
A user having both couchbase and mongo configurations noticed the two clashed because of same bean naming (in this particular case, "customConversions").

PR #93, opened 2016-01-27
DATACOUCH-180 - Allow named parameters in inline N1QL queries

simonbasle:
@daschl can you review this? :bow:

### DIFF
--- a/src/integration/java/org/springframework/data/couchbase/repository/N1qlPlaceholderTests.java
+++ b/src/integration/java/org/springframework/data/couchbase/repository/N1qlPlaceholderTests.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2013-2015 the original author or authors.
+ *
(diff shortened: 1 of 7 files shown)


daschl:
one minor comment, otherwise :+1:

simonbasle:
I wanted to limit the syntax for placeholders to syntaxes that are native to N1QL but still make sense in the Spring Data world.

Both `$1` and `$name` syntaxes are native to N1QL and AFAIK accepted (although maybe not canonical) syntaxes in Spring Data. The `$name` would need a `@Param("name")` annotation.

@olivergierke can you confirm this looks good to you as well?

simonbasle:
updated the PR, detection of the placeholder type is now done by analyzing the query and mixing named and positional placeholders is now throwing an `IllegalArgumentException` at repository creation.

daschl:
:+1:

PR #94, opened 2016-01-29
DATACOUCH-196 - Fix ViewIndexed generated map and improve usability

simonbasle:
The generated view's default map function now emits meta.id so that CRUD findAll(ids) works.

The default value for the view name is now "all".

The ViewIndexed isn't documented as applicable to METHODs (deferred feature).


simonbasle:
pinging @daschl for review :8ball:

daschl:
good idea! :+1:

PR #95, opened 2016-02-05
DATACOUCH-199 - Make automatic index creation opt-in

simonbasle:
In addition to annotating the repository with xxxIndexed annotations, user must now opt-in to the feature by redefining the indexManager bean in the configuration. This is so production use of this feature is actively discouraged.


simonbasle:
@daschl or @olivergierke, please review this (before 2.0 GA can be kicked off the door :dancers:)

daschl:
I think this is a good change, :+1:

PR #96, opened 2016-02-20
DATACOUCH-59 | Configuration for updating expire settings for documen…

awislowski:
…ts when read is made. Configuration is based on Document annotation. When read of document takes place then touch action is executed on document. It is an asynchronous action so it does not block read.


simonbasle:
I'm really sorry, 2.0.0 has now been released so branch `2.0.x` is the maintenance branch for that version now. This change would rather make 2.1.0, so the PR needs to be re-made against `master` :(

The good news is it should be as straightforward as closing this PR and reopening one on master with the same fork/branch...

You can update your fork and rebase your branch on master if you want but I think git should still be able to figure it out without doing that...

PR #97, opened 2016-02-22
DATACOUCH-207 - Makes all BeanNames constants public

simonbasle:
This makes all bean names public constants, which can be used in Conditional annotation in Spring Boot for instance (cc @snicoll).


snicoll:
LGTM. Maybe the Javadoc of `BeanNames` could be refined now it is public?

simonbasle:
updated the javadoc

snicoll:
:+1:

PR #98, opened 2016-02-22
DATACOUCH-59 | Configuration for updating expire settings for documen…

awislowski:
…ts when read is made. Configuration is based on Document annotation. When read of document takes place then touch action is executed on document. It is an asynchronous action so it does not block read.

due to @simonbasle request in https://github.com/spring-projects/spring-data-couchbase/pull/96


simonbasle:
Cherry-picked (with minor javadoc edits) as 40e4212
Thanks @awislowski !

PR #99, opened 2016-02-22
DATACOUCH-209 - Change default Consistency to READ_YOUR_OWN_WRITES

simonbasle:
Formerly UPDATE_AFTER, but the inconsistent view this would give eg. while writing tests or exploring Spring Data Couchbase was confusing to users.

cc. @snicoll for review, related to spring-projects/spring-boot#5159


snicoll:
LGTM

PR #100, opened 2016-02-23
DATACOUCH-204 - Add package-info on all packages

simonbasle:
ping @olivergierke for review


olivergierke:
Two minor things: I don't think we need Apache license headers for packages. Make sure you add the newlines at the EOF that are currently missing. LGTM beyond that.

simonbasle:
updated, I thought I had removed these headers :sweat_smile:

PR #103, opened 2016-03-07
Introduce CouchbaseConfigurer

snicoll:
Rather than having to extend a few methods from the base Configuration
class, this commit introduces a strategy interface to gather them.

`CouchbaseConfigurer` represents all the components that the Spring Data
integration requires.


simonbasle:
:+1: merged, I'll polish a bit with tests and more javadoc in the spring-projects:draft/separateConfigConcerns branch

PR #104, opened 2016-03-09
DATACOUCH-215 - Separate config concerns between SDK core and Spring Data

simonbasle:
The aim here is to facilitate integration with Spring Boot while allowing apps to have a dependency on the Couchbase SDK (`Bucket` and `Cluster`) without necessarily pulling in Spring Data Couchbase.

Pinging @daschl @snicoll @olivergierke for quick review (as per the ongoing email thread).


simonbasle:
see issue [DATACOUCH-215](https://jira.spring.io/browse/DATACOUCH-215)

PR #102, opened 2016-03-03
DATACOUCH-211 - Applying sort that's passed in via the Pagenate object

longwatertech:

PR #105, opened 2016-03-09
DATACOUCH-214 - Allow case insensitive ordering via Sort.Order class

longwatertech:


src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
```diff
@@ -145,8 +146,11 @@ public static String getDottedPathWithAlternativeFieldNames(PersistentPropertyPa
     List<com.couchbase.client.java.query.dsl.Sort> cbSortList = new ArrayList<com.couchbase.client.java.query.dsl.Sort>();
     for (Sort.Order order : sort) {
       String orderProperty = order.getProperty();
-        //FIXME the order property should be converted to its corresponding fieldName
+      //FIXME the order property should be converted to its corresponding fieldName
       Expression orderFieldName = i(orderProperty);
+      if (order.isIgnoreCase()) {
+        orderFieldName = lower(orderFieldName);
```
simonbasle: in N1QL `LOWER` on anything but a String will return null. You could use the `TOSTRING` function as well, which is accessible through `TypeFunctions.toString(Expression x)` in the DSL

simonbasle:
Hi @longwatertech and thanks for the PR :gift:
As I commented inline, you should combine `LOWER` and `TOSTRING` for maximum effect. Can you make the modification and squash the 3 commits into a single commit?

Also I'll need you to confirm that you have signed Spring's [CLA](https://support.springsource.com/spring_committer_signup) as stated in the [contributing instructions](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

If you can do both by tomorrow, this should be able to make it to `2.1 Hopper RC1` :smiley:

longwatertech:
Hi @simonbasle - thanks for the feedback. I'll get the `TOSTRING` change added and squash those commits. Regarding the CLA I have two questions: 1) is the project 'Spring Data Access'? 2) Who is the project lead? Thanks, James.

simonbasle:
@longwatertech the project is "Spring Data" I think (although I'm not entirely sure). Project lead is Oliver Gierke ( @olivergierke).

longwatertech:
@simonbasle - also, I signed the CLA, but the closest project I could see in the list was 'Spring Data Access' so I chose that one. Let me know if I need to select another. Thanks.

simonbasle:
sounds good to me :+1:

simonbasle:
you can now work on the second one, #102 :wink:

PR #106, opened 2016-03-10
DATACOUCH-216 - Add template getter on repository interface

simonbasle:


simonbasle:
pretty trivial change so I'm going to :+1:

PR #107, opened 2016-03-10
DATACOUCH-212 - Take Version/CAS into account on save

simonbasle:
@daschl I'm interested in what you think about this one


daschl:
like the idea, can you flesh out the commit msg a bit more? :)

PR #109, opened 2016-03-23
DATACOUCH-212: Add failing test case for multiple concurrent saves of…

Aloren:
… new document

### DIFF
--- a/src/integration/java/org/springframework/data/couchbase/core/CouchbaseTemplateTests.java
+++ b/src/integration/java/org/springframework/data/couchbase/core/CouchbaseTemplateTests.java
@@ -26,12 +26,20 @@ import static org.junit.Assert.*;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 import com.couchbase.client.java.Bucket;
 import com.couchbase.client.java.document.RawJsonDocument;
@@ -40,12 +48,13 @@ import com.couchbase.client.java.query.N1qlParams;
 import com.couchbase.client.java.query.N1qlQuery;
 import com.couchbase.client.java.query.N1qlQueryResult;
 import com.couchbase.client.java.query.consistency.ScanConsistency;
-import com.couchbase.client.java.query.dsl.Expression;
 import com.couchbase.client.java.view.Stale;
 import com.couchbase.client.java.view.ViewQuery;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TestName;
 import org.junit.runner.RunWith;
 
 import org.springframework.beans.factory.annotation.Autowired;
@@ -68,6 +77,9 @@ import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
 @ContextConfiguration(classes = IntegrationTestApplicationConfig.class)
 @TestExecutionListeners(CouchbaseTemplateViewListener.class)
 public class CouchbaseTemplateTests {
+
+	@Rule
+	public TestName testName = new TestName();
 
 	@Autowired
 	private Bucket client;
@@ -132,7 +144,11 @@ public class CouchbaseTemplateTests {
 		assertEquals("Mr. A", resultConv.get("name"));
 
 		doc = new SimplePerson(id, "Mr. B");
-		template.insert(doc);
+		try {
+			template.insert(doc);
+		} catch (OptimisticLockingFailureException e) {
+			//ignore, since this insert should fail
+		}
 
 		resultDoc = client.get(id, RawJsonDocument.class);
 		assertNotNull(resultDoc);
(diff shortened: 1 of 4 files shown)


simonbasle:
As noted in the issue, this is because the items are still with a 0-valued version attribute, so the template assumes that version checking is not to be enforced. what you could do is do a first insert, extract the `version` field and use it when trying to save the items in parallel. Something like:

```java
@Test
  public void shouldFailOnMultipleConcurrentSaves() throws Exception {
    final String key = "key";
    final AtomicLong counter = new AtomicLong();
    final AtomicLong optimisticLockCounter = new AtomicLong();
    ExecutorService pool = Executors.newFixedThreadPool(5);
    VersionedData initialData = new VersionedData(key, "value-initial");
    versionedDataRepository.save(initialData);
    assertNotEquals(0L, initialData.version);
    final long cas = initialData.version;

    Callable<Void> task = new Callable<Void>() {
        @Override
        public Void call() throws Exception {
            long counterValue = counter.incrementAndGet();
            String data = "value-" + counterValue;
            VersionedData messageData = new VersionedData(key, data);
            messageData.version = cas;
            try {
              versionedDataRepository.save(messageData);
            } catch (OptimisticLockingFailureException e) {
              optimisticLockCounter.incrementAndGet();
            }
            //should save operation throw OptimisticLockingFailureException on next attempts to save?
            return null;
        }
    };
    Collection<Callable<Void>> tasks = Collections.nCopies(5, task);

    List<Future<Void>> futures = pool.invokeAll(tasks);
    for (Future future : futures) {
        future.get(5, TimeUnit.SECONDS);
    }

    assertEquals(4, optimisticLockCounter.intValue());
    assertEquals("value-1", versionedDataRepository.findOne("key").data);
  }
```

simonbasle:
actually the value itself isn't very relevant as any of the 5 threads could win the race to mutate the doc. Only the fact that 1 mutation happened and 4 optimistic locking exceptions happened is really relevant (edited above)

Aloren:
Yes, you are correct, my bad. Fixed my test according to your comment. But your version of test assumes that document is present in couchbase before we concurrently try to update it. In my test document does not exist and we concurrently try to create it in couchbase. I assume that first attempt should succeed and other 4 should fail with some kind of exception, but it does not happen. So each save attempt overwrites previous. This does not allow to implement merge properly.

Aloren:
@simonbasle I have updated pull request with fast fix. Please review. Thanks.

Aloren:
I have also found an issue with CouchbaseTemplate insert method. Only first attempt of insert should succeed, next attempts should fail with OptiomisticLockingFailureException (as couchbase client throws DocumentAlreadyExistsException).

simonbasle:
@olivergierke I'd like your input on this :) did we go a wrong direction at some point? what should be the behavior of multiple subsequent repository saves, depending on the `@Version` property?

vasilievip:
@simonbasle any updates, can you apply changes? @olivergierke seems do not care which direction you went.

src/integration/java/org/springframework/data/couchbase/repository/SimpleCouchbaseRepositoryTests.java
```diff
@@ -220,6 +223,40 @@ public void shouldTakeVersionIntoAccountWhenDoingMultipleUpdates() {
...
+      }
+    };
+    Collection<Callable<Void>> tasks = Collections.nCopies(5, task);
+
+    List<Future<Void>> futures = pool.invokeAll(tasks);
+    for (Future future : futures) {
+      future.get(5, TimeUnit.SECONDS);
+    }
+
+    versionedDataRepository.delete(key);
```
simonbasle: shouldn't the delete be made at the beginning of test? maybe test both cases (one where a pre-existing version of the doc exists => update, the other where it is `deleteIfExist` => insert)?

simonbasle:
after coming back to this, it actually looks good. @Aloren since DATACOUCH-212 has shipped, I have opened a follow-up issue: [DATACOUCH-224](https://jira.spring.io/browse/DATACOUCH-224).
Can you look at my last comment in this PR, do a last modification if you agree and then squash all the changes together into a single commit, with title "`DATACOUCH-224 - Improve concurrent save/insert, optimistic locking`"?

Aloren:
Hi @simonbasle !
I will try to do the fix today. Thanks

simonbasle:
awesome @Aloren, looking forward to it ;)

Aloren:
Please check if my commit includes all the needed changes. Thanks

src/integration/java/org/springframework/data/couchbase/core/CouchbaseTemplateTests.java
```diff
@@ -402,6 +422,67 @@ public void shouldLoadVersionPropertyOnFind() throws Exception {
...
+				boolean saved = false;
+				while(!saved) {
+					long counterValue = counter.incrementAndGet();
+					VersionedClass messageData = template.findById(key, VersionedClass.class);
+					messageData.field = "value-" + counterValue;
+					try {
+						template.save(messageData);
+						saved = true;
+					} catch (OptimisticLockingFailureException e) {
+					}
```
simonbasle: maybe use a counter here too, to assert that 1 out of 3 updates succeeded?

src/integration/java/org/springframework/data/couchbase/repository/SimpleCouchbaseRepositoryTests.java
```diff
@@ -220,6 +238,66 @@ public void shouldTakeVersionIntoAccountWhenDoingMultipleUpdates() {
...
+        boolean saved = false;
+        while(!saved) {
+          long counterValue = counter.incrementAndGet();
+          VersionedData messageData = versionedDataRepository.findOne(key);
+          messageData.data = "value-" + counterValue;
+          try {
+            versionedDataRepository.save(messageData);
+            saved = true;
+          } catch (OptimisticLockingFailureException e) {
+          }
```
simonbasle: maybe add a counter here too, to assert that exactly 1 out of 5 concurrent updates succeeded?

simonbasle:
@Aloren just one small nitpick on the "update" test cases, and this will be perfect ;) (see comments)

src/integration/java/org/springframework/data/couchbase/core/CouchbaseTemplateTests.java
Aloren: all saves will succeed here since all of them are retried with updated version. Maybe you are talking about another test case?

simonbasle: you are right, indeed. went too fast in reading this particular test :)

simonbasle:
Looks good, merged as commit 811c024 👍

simonbasle:
> just a side note @Aloren looks like your work address is not associated to your github account (so the commits don't appear linked to you github identity)

Aloren:
Thank you! (Commited under my work account :) Hope, this is not an issue :) )

simonbasle:
not an issue at all, but you should be able to add your work address to your github account's emails and it will properly match your commits 😄

PR #112, opened 2016-05-17
DATACOUCH-226 

geoffreymina:
Greetings,
This patch solves a problem I was having with circular references.  Our use case was pulling some complex objects from MySQL via Hibernate that had OneToMany relationships using the "mappedby" which requires circular references.  

This change modifies the MappingCouchbaseConverter to (optionally) be strict about the @Field annotation.  If setEnableStrictFieldChecking(true) is called, the mapper will only process properties that are annotated and ignore non-annotated properties entirely.

I have also submitted test coverage to show that the expected behavior is present with the boolean in both the true and false state.

Thanks for your consideration.
Geoff Mina

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
@@ -16,6 +16,13 @@
 
 package org.springframework.data.couchbase.core.convert;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
 import org.springframework.core.CollectionFactory;
@@ -45,13 +52,6 @@ import org.springframework.data.util.ClassTypeInformation;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
 
 /**
  * A mapping converter for Couchbase.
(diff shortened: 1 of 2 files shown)


simonbasle:
hey @geoffreymina, I'm a bit late on getting back to you but I thought of something: have you tried the `@Transient` annotation from the `Spring Data` framework? Sounds like your problem is the kind of thing targeted by the annotation.

If it works, then you won't have to wait for a release so I think that'd be a huge win 😃

Note that `@Field` is couchbase-specific and was introduced in the "dumbed-down" repository model that is experimental in the native Couchbase SDK, essentially to give an alias to an attribute (change the name that is used in the stored JSON).

geoffreymina:
@simonbasle The MappingCouchbaseConverter doesn't take any annotations into account when recursively converting the POJO into a CouchbaseDocument.  That being said - I felt since this was a self-contained Couchbase specific operation, it made sense to actually use the @Field annotation.  I suppose a @Transient annotation could be treated equally in this scenario.

Either way - no current annotation has any impact on preventing infinite recursion with circular references, so a patch would be required regardless.  If you feel strongly about the use of the @Transient annotation instead of @Field - i can certainly make that change.

Thanks,
Geoff

simonbasle:
I just launched your test, replacing the `@Field` with a `@Transient`. The test passes (or rather it fails in the part where it assert the property is serialized in non-strict mode), so that made me think it was a good alternative. Transient is supposed to make it so that the annotated attribute doesn't even show up as a `PersistentProperty` during conversion...

If your own usage shows that it doesn't work in your production app, then the test should probably be modified to better reflect the problem, and there is something more fishy going on... Otherwise, as I said, win 😉

geoffreymina:
OK - i see where that is implemented.  That still doesn't solve my problem as I cannot annotate these properties @Transient because this object is also being utilized by Hibernate.  Our pattern is that we pull data from MySQL (JPA/Hibernate) and cache it in Couchbase directly.  I still need the "strictFieldChecking" to prevent needing two different objects as I have the properties annotated with both @Field and @Column/@OneToMany/@ManyToOne, etc  - except on the circular references, where we leave out the @Field to prevent the StackOverflowError.

I have a working patch in our application, so it isn't the end of the world if this pull request is denied.  Would be nice to not have to maintain our own branch though :)

Thanks for the consideration.
Geoff

simonbasle:
ah I see, yeah this makes sense as the `@Transient` would impact Spring Data Hibernate as well.

simonbasle:
I'll do a few cleanup (split the test in two and fix a few minor formatting), squash the changes and merge this in then ;)

PR #113, opened 2016-05-25
DATACOUCH-228 - Make findByView use async client internally

simonbasle:
This brings better performance when querying large views. Instead of
looping on a synchronously acquired list of documents to produce a list
of entities, the async API can be used internally to:
 - query the view
 - fetch the documents
 - map each document to an entity
 - collect into a List

One can then block on that Observable to get and return the List as
usual.

Perf gain has been validated by one of our users. @daschl I'll need a quick review on this one :)


daschl:
👍

PR #114, opened 2016-05-25
DATACOUCH-229 - Check N1QL error codes rather than message

simonbasle:
This should avoid problems when the message changes on the server side
(like in Couchbase Server 4.5).


simonbasle:
this was a bad cherry pick and was cancelled, commit 3c92335 has been forcibly removed  from master.

PR #115, opened 2016-05-25
DATACOUCH-229 - Check N1QL error codes rather than message

simonbasle:
This should avoid problems when the message changes on the server side
(like in Couchbase Server 4.5).

@daschl pretty simple change, might remove the message check altogether
but I'm waiting for a confirmation about the code from query team.


daschl:
👍

simonbasle:
looking at the query service code, 4300 only happens for index not found :+1:

PR #117, opened 2016-06-15
DATACOUCH-235 - Fixed multiple Couchbase environments.

simonbland:
Defining `<couchbase:env/>` in [XML configuration](http://docs.spring.io/spring-data/couchbase/docs/2.0.0.RELEASE/reference/html/#configuration-xml) was causing a double
instantiation of DefaultCoreEnvironment, with subsequent warning in
logs:

```
[main] WARN c.c.client.core.env.CoreEnvironment - More than 1 Couchbase
Environments found (2), this can have severe impact on performance and
stability. Reuse environments!
```

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/config/CouchbaseSingleEnvironmentParserTest.java
+++ b/src/test/java/org/springframework/data/couchbase/config/CouchbaseSingleEnvironmentParserTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2012-2016 the original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.couchbase.config;
+
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+import org.junit.Test;
+import org.springframework.beans.factory.support.BeanDefinitionReader;
+import org.springframework.beans.factory.support.DefaultListableBeanFactory;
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/data/couchbase/config/CouchbaseSingleEnvironmentParserTest.java
```diff
@@ -0,0 +1,68 @@
...
+    context.refresh();
+    CouchbaseEnvironment env = context.getBean("singleEnv", CouchbaseEnvironment.class);
+    context.close();
+    
+    System.out.flush();
+    System.setOut(oldSystemOut);
+    System.out.println(logs.toString());
+    
+    assertThat(env, is(instanceOf(DefaultCouchbaseEnvironment.class)));
+    assertTrue(logs.toString().indexOf("More than 1 Couchbase Environments found (2)") == -1);
```
simonbasle: very minor, but maybe don't assert the "(2)" part?

simonbasle:
Outstanding! Congrats on your first PR that is perfectly following the project's style 🎉🎈
The change is exactly what I had in mind 👍 

1 minor comment. If you decide in favor of it, be sure to do a `commit --amend`of the change 😉

simonbasle:
@simonbland can you also confirm that you've signed the [Contributor's Licence Agreement (CLA)](https://support.springsource.com/spring_committer_signup)?

simonbland:
Thank you @simonbasle :blush:

I've modified `CouchbaseSingleEnvironmentParserTest.java` as you suggested and amended the commit.

I also confirm that I've signed the CLA :wink:

simonbasle:
Ah one last thing! The test turns out is a bit brittle. Running this along with other unit tests fails, because other tests have already created their own environment. I think I'll change the SDK at some point to 1) allow to get the instanceCount value / get another form of notification on multiple instance checks and 2) make it so that the count is reduced by one each time an `Environment` is shut down...

In the meantime, for the test maybe the simplest solution would be to replace the log capturing by a retrieval of the `instanceCounter` via reflection?

you can use this snippet before and after parsing the context:
```java
(Integer) ReflectionTestUtils.getField(DefaultCoreEnvironment.class, "instanceCounter");
```
then at the end, assert like so:

```java
assertThat(instanceCounterAfter, is(instanceCounterBefore + 1));
```

simonbland:
I agree, this is a better way to test. I wasn't really happy with this logs parsing solution anyway.

I've replaced logs capturing with probing of `instanceCounter` through reflection, as you proposed, and amended the commit again.

`org.springframework.test.util.ReflectionTestUtil` is nice indeed!

Thanks 😉

simonbasle:
Verified and merged, thanks again @simonbland for the 🎁 of PR :wink:

simonbland:
My pleasure :wink:

Thank you too for your help 👍

PR #122, opened 2016-09-24
DATACOUCH-246 Projections support for query derivation

subalakr:
Thank you for proposing a pull request. This template will guide you through the essential steps necessary for a pull request.
Make sure that:
- [x] You provide your full name and an email address registered with your GitHub account. If you’re a first-time submitter, make sure you have completed the [Contributor’s License Agreement form](https://support.springsource.com/spring_committer_signup).

Used the result processor projecting information to just fetch the selected
fields and again to post process results to projection object/interfaces.
Added a DTOInstantiating converter for DTO similar to previous implementations
of projections by Oliver.
Added tests for projections using DTO and interface.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
-import com.couchbase.client.core.lang.Tuple2;
 import com.couchbase.client.java.document.json.JsonArray;
 import com.couchbase.client.java.document.json.JsonObject;
 import com.couchbase.client.java.document.json.JsonValue;
@@ -40,12 +39,17 @@ import org.springframework.data.repository.query.ParameterAccessor;
 import org.springframework.data.repository.query.ParametersParameterAccessor;
 import org.springframework.data.repository.query.QueryMethod;
 import org.springframework.data.repository.query.RepositoryQuery;
+import org.springframework.data.repository.query.ResultProcessor;
+import org.springframework.data.repository.query.ReturnedType;
 import org.springframework.data.util.StreamUtils;
 import org.springframework.util.Assert;
 
 /**
  * Abstract base for all Couchbase {@link RepositoryQuery}. It is in charge of inspecting the parameters
  * and choosing the correct {@link N1qlQuery} implementation to use.
+ *
+ * @author Simon Baslé
+ * @author Subhashni Balakrishnan
  */
 public abstract class AbstractN1qlBasedQuery implements RepositoryQuery {
 
@@ -53,6 +57,7 @@ public abstract class AbstractN1qlBasedQuery implements RepositoryQuery {
 
   protected final CouchbaseQueryMethod queryMethod;
   private final CouchbaseOperations couchbaseOperations;
+  private Class<?> returnedType;
 
   protected AbstractN1qlBasedQuery(CouchbaseQueryMethod queryMethod, CouchbaseOperations couchbaseOperations) {
     this.queryMethod = queryMethod;
@@ -61,6 +66,7 @@ public abstract class AbstractN1qlBasedQuery implements RepositoryQuery {
 
   /**
    * The statement for a count() query. This must aggregate using count with the alias {@link CountFragment#COUNT_ALIAS}.
+   *
    * @see CountFragment
    */
   protected abstract Statement getCount(ParameterAccessor accessor, Object[] runtimeParameters);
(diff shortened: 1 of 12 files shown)


subalakr:
@daschl @olivergierke It would be great if you can review. It is my first changes with spring so please bear with me if I did something very wrong.

(Regarding cla, it is not yet done. Couchbase is looking at a corporate cla)

ingenthr:
Apologies for the delay in CLA, that's with me and we're getting it reviewed now.

src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
```diff
@@ -75,14 +86,40 @@ public static Expression escapedBucket(String bucketName) {
...
+  public static FromPath createSelectClauseForEntity(String bucketName, ReturnedType returnedType) {
     Expression bucket = escapedBucket(bucketName);
     Expression metaId = path(meta(bucket), "id").as(CouchbaseOperations.SELECT_ID);
     Expression metaCas = path(meta(bucket), "cas").as(CouchbaseOperations.SELECT_CAS);
+    List<Expression> expList = new ArrayList<Expression>();
+    expList.add(metaId);
+    expList.add(metaCas);
+
+    if (returnedType.isProjecting()) {
+      List<String> properties = returnedType.getInputProperties();
```
olivergierke: I guess this could be simplified to:

```java
CouchbasePersistentEntity<?> entity = mappingContext.getPersistentEntity(returnedType.getDomainType());

for (String  property : properties) {
  expList.add(path(bucket, entity.getPersistentProperty(property).getFieldName()));
}
```

Creating a `MappingContext` from scratch is suboptimal here. Is there a way we can reuse an existing instance already available? Generally speaking I wonder why the execution here has to do the property-to-field-name-mapping. Shouldn't that be the responsibility of the query execution in `CouchbaseTemplate`?

src/test/java/org/springframework/data/couchbase/core/Beer.java
```diff
@@ -16,6 +16,7 @@
 
 package org.springframework.data.couchbase.core;
 
+import org.springframework.beans.factory.access.BootstrapException;
```
olivergierke: That shouldn't be necessary, should it?

src/test/java/org/springframework/data/couchbase/repository/query/PartTreeN1qBasedQueryTest.java
```diff
@@ -117,12 +126,129 @@ public void testGetCountExcludesDynamicSortClause() throws Exception {
 
 	}
 
+	@Test
```
olivergierke: That looks like a lot of mocking going on here. Should this rather be an integration test?

To avoid the `ProjectionInformation` implementation, simply use `SpelAwareProxyProjectionFactory` and call `getProjectionInformation(…)` on it.

src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
```diff
@@ -71,14 +78,22 @@ protected AbstractN1qlBasedQuery(CouchbaseQueryMethod queryMethod, CouchbaseOper
...
-  protected abstract Statement getStatement(ParameterAccessor accessor, Object[] runtimeParameters);
+  protected abstract Statement getStatement(ParameterAccessor accessor, Object[] runtimeParameters, ReturnedType returnedType);
 
   protected abstract JsonValue getPlaceholderValues(ParameterAccessor accessor);
 
   @Override
   public Object execute(Object[] parameters) {
     ParametersParameterAccessor accessor = new ParametersParameterAccessor(queryMethod.getParameters(), parameters);
-    Statement statement = getStatement(accessor, parameters);
+
```
olivergierke: This looks like we're issuing a query and then basically map the results back onto a DTO. Is there a way we can instruct the query to let the query execution return the requested type in the first place?

src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
olivergierke: I think the best way to obtain a `MappingContext` is `couchbaseOperations.getConverter().getMappingContext()`.

src/test/java/org/springframework/data/couchbase/core/Beer.java
subalakr: Not needed, fixed it

src/test/java/org/springframework/data/couchbase/repository/query/PartTreeN1qBasedQueryTest.java
subalakr: Thanks for pointing out SpelAwareProxyProjectionFactory. The mocking is bit less now, I think we should move it to integration test next release?

src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
subalakr: I tried moving all of the mapping to DTO to couchbaseOperations but due interface mapping, it still happens on AbstractN1QLBasedQuery because I couldn't pass down the domain type and interface return type until findByN1QL method without breaking api, we'll discuss more on this tomorrow with Simon

subalakr:
Merged closing

ingenthr:
Nice work @subalakr !

PR #123, opened 2016-11-23
DATACOUCH-251 Wrap base where criteria with paranthesis

subalakr:
Thank you for proposing a pull request. This template will guide you through the essential steps necessary for a pull request.
Make sure that:

- [x] You provide your full name and an email address registered with your GitHub account. If you’re a first-time submitter, make sure you have completed the [Contributor’s License Agreement form](https://support.springsource.com/spring_committer_signup).

Base criteria for the query should be wrapped before adding the entity filter
criteria so the priority of filtering is clear

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
@@ -20,6 +20,7 @@ import static com.couchbase.client.java.query.Select.select;
 import static com.couchbase.client.java.query.dsl.Expression.i;
 import static com.couchbase.client.java.query.dsl.Expression.path;
 import static com.couchbase.client.java.query.dsl.Expression.s;
+import static com.couchbase.client.java.query.dsl.Expression.x;
 import static com.couchbase.client.java.query.dsl.functions.AggregateFunctions.count;
 import static com.couchbase.client.java.query.dsl.functions.MetaFunctions.meta;
 import static com.couchbase.client.java.query.dsl.functions.StringFunctions.lower;
@@ -114,7 +115,7 @@ public class N1qlUtils {
     if (baseWhereCriteria == null) {
       baseWhereCriteria = typeSelector;
     } else {
-      baseWhereCriteria = baseWhereCriteria.and(typeSelector);
+      baseWhereCriteria = x("(" + baseWhereCriteria.toString() + ")").and(typeSelector);
     }
     return baseWhereCriteria;
   }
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
```diff
@@ -114,7 +115,7 @@ public static Expression createWhereFilterForEntity(Expression baseWhereCriteria
     if (baseWhereCriteria == null) {
       baseWhereCriteria = typeSelector;
     } else {
-      baseWhereCriteria = baseWhereCriteria.and(typeSelector);
+      baseWhereCriteria = x("(" + baseWhereCriteria.toString() + ")").and(typeSelector);
```
simonbasle: no better way of doing this, maybe `Expression.prefix/suffix/infix` could be made public in a future SDK release?
you could add a test that uses a `baseWhereCriteria` to `N1qlUtilsTest`:

```java
  @Test
  public void testCreateWhereFilterForEntityWithBaseCriteria() throws Exception {
    String expected = "(field >= 30 OR field2 = \"foo\") AND `_class` = \"java.lang.String\"";
    CouchbaseConverter converter = mock(CouchbaseConverter.class);
    when(converter.getTypeKey()).thenReturn("_class");
    EntityMetadata metadata = mock(EntityMetadata.class);
    when(metadata.getJavaType()).thenReturn(String.class);

    String real = N1qlUtils.createWhereFilterForEntity(
            x("field").gte(30).or(x("field2").eq(s("foo"))),
            converter, metadata).toString();

    assertEquals(expected, real);
  }
```

src/test/java/org/springframework/data/couchbase/repository/query/PartTreeN1qBasedQueryTest.java
```diff
@@ -112,17 +114,50 @@ public void testGetCountExcludesDynamicSortClause() throws Exception {
...
+				+ " AND `_class` = \"org.springframework.data.couchbase.core.Beer\"", statement.toString());
+	}
+
 	public static interface TestRepository extends CrudRepository<Beer, String> {
 
 		Page<Beer> findByNameOrderByName(String name, Pageable pageRequest);
 
 		Page<Beer> findByName(String name, Pageable pageRequest);
 
+		List<Beer> findByIdOrName(String id, String name);
```
simonbasle: I would probably also test in an integration test that that sort of query does indeed return the expected results from db when executed from end to end...

simonbasle:
@subalakr I think you can omit the pull-request template now that you've signed the CLA, since you're the official main committer to the project now 😄

src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
subalakr: yes no other way for now, will expose an api to group expressions

subalakr:
merged

PR #128, opened 2017-01-03
DATACOUCH-265: Support pageable and sort for String based SpEL queries

subalakr:
Pageable and sort parameters were ignored before. This should fix it.


simonbasle:
@subalakr @daschl I think there's a limitation, if the string statement already has an `ORDER BY` or `LIMIT` clause... That should be at least made clear in the documentation that combining a `Sort` or `Pageable` parameter with a string query that already contains such a clause will produce invalid N1QL and be rejected by the server ;)

subalakr:
Thanks Simon :) Added a note in the docs.

prafsoni:
Code, didn't merged to 2.2.x branch, v2.2.9 doesn't support Pageable.
https://jira.spring.io/browse/DATACOUCH-265 mentions 2.2.1 as fix version but, 2.2.x branch code doesn't reflect changes to getStatement method https://github.com/spring-projects/spring-data-couchbase/blob/2.2.x/src/main/java/org/springframework/data/couchbase/repository/query/StringN1qlBasedQuery.java  Line 292. Branch 3.0.x does reflect changes.

PR #130, opened 2017-01-19
DATACOUCH-146 Reactive and RxJava1 repositories support

subalakr:
- Adds experimental support for reactor types in reactive CRUD and Sorting
  repositories
- Adds experimental support for rxjava1 CRUD repository
- Refactor configuration and query classes to be reusable by reactive repositories


src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplate.java
```diff
@@ -0,0 +1,411 @@
...
+    private final TranslationService translationService;
+    protected final MappingContext<? extends CouchbasePersistentEntity<?>, CouchbasePersistentProperty> mappingContext;
+    private Consistency configuredConsistency = Consistency.DEFAULT_CONSISTENCY;
+    private static final WriteResultChecking DEFAULT_WRITE_RESULT_CHECKING = WriteResultChecking.NONE;
+
+    private WriteResultChecking writeResultChecking = DEFAULT_WRITE_RESULT_CHECKING;
+
+
+    @SuppressWarnings("unchecked")
+    public <T> Mono<T> save(T objectToSave) {
```
mp911de: See https://gist.github.com/subalakr/4ea0d1f61e776f2f5cc8c62d946e103b for comments. In particular, providing core reactive support using RxJava1 and using/wrapping types in the `ReactiveCouchbaseTemplate` seems quite appealing. It does not require to maintain the same code twice.

src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplate.java
```diff
@@ -0,0 +1,411 @@
...
+        return ReactiveWrapperConverters.toWrapper(doPersist(objectToSave, PersistTo.NONE, ReplicateTo.NONE), Mono.class);
+    }
+
+    @SuppressWarnings("unchecked")
+    public <T> Flux<T> save(Iterable<T> batchToSave) {
+        return Flux.fromIterable(batchToSave)
```
mp911de: Batch saving pretends some sort of atomic operation but in in fact each item is saved individually. I propose to shift that responsibility to the user code so users can decide upon batch sizes, call `save(T)` on their own, etc.

src/main/java/org/springframework/data/couchbase/core/RxJavaCouchbaseTemplate.java
```diff
@@ -0,0 +1,348 @@
...
+        return Observable.from(batchToRemove)
+                .flatMap(object -> remove(object));
+    }
+
+    public <T> Observable<T> remove(T objectToRemove, PersistTo persistTo, ReplicateTo replicateTo) {
+        return doRemove(objectToRemove, persistTo, replicateTo);
+    }
+
+    public <T> Observable<T> remove(Collection<T> batchToRemove, PersistTo persistTo, ReplicateTo replicateTo) {
+        return Observable.from(batchToRemove)
```
mp911de: See above regarding batch operations.

src/main/java/org/springframework/data/couchbase/core/RxJavaCouchbaseTemplate.java
```diff
@@ -0,0 +1,348 @@
...
+        return Observable.from(batchToSave)
+                .flatMap(object -> save(object, persistTo, replicateTo));
+    }
+
+    public <T> Observable<T> remove(T objectToRemove) {
+        return doRemove(objectToRemove, PersistTo.NONE, ReplicateTo.NONE);
+    }
+
+    public <T> Observable<T> remove(Collection<T> batchToRemove) {
+        return Observable.from(batchToRemove)
```
mp911de: See above regarding batch operations.

src/main/java/org/springframework/data/couchbase/core/RxJavaCouchbaseTemplate.java
```diff
@@ -0,0 +1,348 @@
...
+    protected final MappingContext<? extends CouchbasePersistentEntity<?>, CouchbasePersistentProperty> mappingContext;
+    private Consistency configuredConsistency = Consistency.DEFAULT_CONSISTENCY;
+
+
+    public <T> Observable<T> save(T objectToSave) {
+        return doPersist(objectToSave, PersistTo.NONE, ReplicateTo.NONE);
+    }
+
+    public <T> Observable<T>  save(Iterable<T> batchToSave) {
+        return Observable.from(batchToSave)
```
mp911de: See above regarding batch operations.

src/main/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepository.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.Serializable;
+
+import org.springframework.data.couchbase.core.ReactiveCouchbaseOperations;
+import org.springframework.data.repository.reactive.ReactiveCrudRepository;
+
+/**
+ * @author Subhashni Balakrishnan
+ * @since 3.0
+ */
+public interface ReactiveCouchbaseRepository<T, ID extends Serializable> extends ReactiveCrudRepository<T, ID> {
```
mp911de: If you omit `getCouchbaseOperations()`, then one repository implementation is sufficient. Repository type adoption is handled in Spring Data Commons that adopts the implementation method result into the declared type.

src/main/java/org/springframework/data/couchbase/repository/RxJava1CouchbaseRepository.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @author Subhashni Balakrishnan
+ * @since 3.0
+ */
+public interface RxJava1CouchbaseRepository<T, ID extends Serializable> extends RxJava1CrudRepository<T, ID> {
+    /**
+     * @return a reference to the underlying {@link RxJavaCouchbaseOperations operation template}.
+     */
+    RxJavaCouchbaseOperations getCouchbaseOperations();
```
mp911de: See above.

src/main/java/org/springframework/data/couchbase/repository/config/EnableRxJavaCouchbaseRepositories.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * @author Subhashni Balakrishnan
+ * @since 3.0
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+@Import(RxJavaCouchbaseRepositoriesRegistrar.class)
+public @interface  EnableRxJavaCouchbaseRepositories {
```
mp911de: One `@EnableReactiveRepositories` annotation should be sufficient. The idea behind the repository support is that it basically does not matter which reactive library your user is using because type adoption is handled in the repository layer. The reactive library used by the data module to provide reactive support is another topic that is independent from the user's library choice.

src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
```diff
@@ -0,0 +1,282 @@
...
+     */
+    private ViewMetadataProvider viewMetadataProvider;
+
+    /**
+     * Create a new Repository.
+     *
+     * @param metadata the Metadata for the entity.
+     * @param operations the reference to the reactive template used.
+     */
+    public SimpleReactiveCouchbaseRepository(final CouchbaseEntityInformation<T, String> metadata,
```
mp911de: See above. You should decide on one library to implement your reactive support (RxJava1 seems the natural choice). Adoption into Project Reactor types (and vice versa) is handled by the repository layer in Spring Data Commons.

src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
```diff
@@ -0,0 +1,282 @@
...
+            designDocument = view.designDocument();
+            viewName = view.viewName();
+        }
+
+        return new ResolvedView(designDocument, viewName);
+    }
+
+    @Override
+    public ReactiveCouchbaseOperations getCouchbaseOperations(){
+        return operations;
```
mp911de: That's the only bit which would not be converted when using reactive type adoption from the repository layer.

src/main/java/org/springframework/data/couchbase/repository/support/SimpleRxJavaCouchbaseRepository.java
```diff
@@ -0,0 +1,279 @@
...
+import rx.Completable;
+import rx.Observable;
+import rx.Single;
+
+/**
+ * Reactive repository base implementation for Couchbase.
+ *
+ * @author Subhashni Balakrishnan
+ */
+public class SimpleRxJavaCouchbaseRepository<T, ID extends Serializable> implements RxJava1CouchbaseRepository<T, ID> {
```
mp911de: See above. You should decide on one library to implement your reactive support (RxJava1 seems the natural choice). Adoption into Project Reactor types (and vice versa) is handled by the repository layer in Spring Data Commons.

subalakr:
Thanks @mp911de. I've updated PR to enable only reactive repositories and the implementation uses RxJava1Template. Really appreciate your help on this.

pom.xml
```diff
@@ -84,6 +84,25 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>io.projectreactor</groupId>
+            <artifactId>reactor-core</artifactId>
+            <version>${reactor}</version>
```
subalakr: Fixed

src/main/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepository.java
subalakr: Fixed, using RxJavaCouchbaseOperations

src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
subalakr: I have added the mapping here mapMono and mapFlux wrapper methods, it would be great to know your thoughts.

mp911de: Wow, that was fast. Your changes suggest you want to provide repository support based on Reactive Streams/Project Reactor types. You use/expose on the other side the `RxJava1Template` through the repository API which looks kind of strange. Conversion (adoption) of reactive types has to happen at somewhere. If you want to provide only `RxJava1Template`, then the repository is the place of choice.

mp911de:
Thanks for your update. I missed the `ReactiveCouchbaseTemplate` first, but it might be a better decision in the long term. The Couchbase driver uses RxJava1 and as far as I have understood it's not entirely clear how the driver will evolve. From that point, it's a natural decision to provide only a `RxJava1Template`. Having a RxJava1 Template and a Project Reactor repository implementation looks a bit strange but it might turn out that's the more flexible approach. 

We're considering promoting RxJava2 repository support and make RxJava1 repository support less prominent because of the RxJava type system and its influences (e.g. completing a `Single` without value propagates a `NoSuchElementException`, even into a `Mono`/`Flux`). RxJava1 support became a moving part recently when we made our first experiences with the overall Spring integration. RxJava 2's evolution and the maintenance outlook for RxJava1 (no further development at the time we're going to release Spring Data with reactive support) play into the idea to support RxJava1 but push RxJava2 more. We have no final decision yet, but I wanted to share our considerations with you.

subalakr:
Couchbase SDK might eventually move to RxJava2. It breaks api so we are bit concerned. 
RxJava1 conversion to reactor types is not very straight forward, I ran into some problems where null values are not allowed by reactive streams specification. 
So repository implementation uses the RxJava1template and works around those. I think it is reasonable to have the backdoor to fix common issues instead of letting data-commons to do the conversion.

PR #132, opened 2017-02-05
DATACOUCH-179 : Customizing typeKeyValue when building a N1qlQuery

zakariaeMahla:


pivotal-issuemaster:
@zakariaeMahla Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=132)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=132) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@zakariaeMahla Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=132)!

src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
```diff
@@ -145,7 +146,7 @@ public static Expression createWhereFilterForEntity(Expression baseWhereCriteria
                                                       EntityMetadata<?> entityInformation) {
     //add part that filters on type key
     String typeKey = converter.getTypeKey();
-    String typeValue = entityInformation.getJavaType().getName();
+    String typeValue = getTypeValue(entityInformation);
```
subalakr: This wouldn't really work, unless the documents are correctly serialized with the correct type value. Also, this covers only query derivation for n1ql. String based n1ql, spatial, view queries also use type mapping which are not addressed.

Type values are used in the Converter class to serialize entity to Couchbase Document and also in deserialization.

src/test/java/org/springframework/data/couchbase/repository/query/support/N1qlUtilsTest.java
```diff
@@ -148,4 +151,18 @@ public void testCreateWhereFilterForEntityWithBaseCriteria() throws Exception {
...
+    CouchbaseConverter converter = mock(CouchbaseConverter.class);
+    when(converter.getTypeKey()).thenReturn("_class");
+    EntityMetadata metadata = mock(EntityMetadata.class);
+    @TypeAlias("CustomDiscriminator") class EntityMock {};
+    when(metadata.getJavaType()).thenReturn(EntityMock.class);
+
+    String real = N1qlUtils.createWhereFilterForEntity(null, converter, metadata).toString();
+
+    assertEquals(expected, real);
+  }
```
subalakr: This particular change needs to be more integration tested. It is also great to have unit tests for more coverage.

src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
zakariaeMahla: I didn't understand when you said : "unless the documents are correctly serialized with the correct type value"?
Actually, the method I have created covers the 2 cases : 
* When @TypeAlias was used when serializing an object. It would be then logical to be also used when deserializing.
* Else, we'll use the previous mapping, ie, the QFN of the class.

For the other String, Spatial and view queries, I'll check.

src/test/java/org/springframework/data/couchbase/repository/query/support/N1qlUtilsTest.java
zakariaeMahla: I added this test for the **if** branch of my private method `N1qlUtils.getTypeValue(EntityMetadata<?> entityInformation)`. The **else** branch (which) has been already tests by the preexisting unit tests

simonbasle:
I'll defer to @subalakr on this one but the feature is definitely worth having, and direction taken for implementation seem rather correct.

Since string-based queries are constructed using `N1qlUtils` as well, it might be transparent for that part. Still please ensure this is working correctly with a simple string-based test that uses an annotated entity ;)

subalakr:
Thanks for the continued effort here @zakariaeMahla 

When entities are saved through repository, they get converted to CouchbaseDocument using a MappingCouchbaseConverter which uses the type value  here
https://github.com/spring-projects/spring-data-couchbase/blob/master/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java#L372

and reads also it. DefaultCouchbaseTypeMapper has a custom TypeAliasAccessor.
https://github.com/spring-projects/spring-data-couchbase/blob/master/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java#L302

https://github.com/spring-projects/spring-data-couchbase/blob/master/src/main/java/org/springframework/data/couchbase/core/convert/DefaultCouchbaseTypeMapper.java

The mapping information for a map function is quite a requirement for view/spatial queries to work correctly
https://github.com/spring-projects/spring-data-couchbase/blob/7ca165d93ea658933c11481840588a9f2db545a2/src/main/java/org/springframework/data/couchbase/repository/support/IndexManager.java#L234

An example of integration test is here, we would like to test against couchbase server and make sure the document can be queried through n1ql, view, spatial
https://github.com/spring-projects/spring-data-couchbase/blob/master/src/integration/java/org/springframework/data/couchbase/repository/Party.java
https://github.com/spring-projects/spring-data-couchbase/blob/master/src/integration/java/org/springframework/data/couchbase/repository/N1qlCouchbaseRepositoryTests.java

MLabusquiere:
#152

davidkelly:
Lets resurrect #152, based on the current master, and close this.

davidkelly:
closing this as there are other PRs which address same

PR #133, opened 2017-02-06
DATACOUCH-279 - AbstractCouchbaseDataConfiguration no longer exists

subalakr:
Motivation
----------
AbstractCouchbaseDataConfiguration was removed and
AbstractCouchbaseConfiguration was made the only configuration support
which implemented CouchbaseConfigurer. This broke integration with
Spring boot as they used CouchbaseConfigurer to auto configure client
initialization properties.

Changes
-------
Bring back support for AbstractCouchbaseDataConfiguration which allows the use
of CouchbaseConfigurer. Also added a similar configuration support for reactive
data repositories. Common beans between reactive and blocking are implemented
in CouchbaseConfigurationSupport class. The existing tests in
AbstractCouchbaseDataConfigurationTest does test this scenario. However the
previous commit had also changed the test because of the unawareness of spring
boot integration.

Results
-------
It is now possible to configure SDC with CouchbaseConfigurer instead of
only having to extend from AbstractCouchbaseConfiguration.

### DIFF
--- a/src/integration/java/org/springframework/data/couchbase/config/AbstractCouchbaseDataConfigurationTest.java
+++ b/src/integration/java/org/springframework/data/couchbase/config/AbstractCouchbaseDataConfigurationTest.java
@@ -16,8 +16,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.data.annotation.Id;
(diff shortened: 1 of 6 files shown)


daschl:
relates to https://github.com/spring-projects/spring-boot/issues/8200

simonbasle:
I'll merge this into `3.0.x` so that the Spring Boot team can reactivate Couchbase support. cc @snicoll @subalakr

PR #139, opened 2017-04-10
DATACOUCH-90 Add commons Jsr310 converters to support Java 8 date/time

subalakr:


src/test/java/org/springframework/data/couchbase/core/mapping/MappingCouchbaseConverterTests.java
```diff
@@ -33,7 +33,9 @@
 import java.util.TreeMap;
 import java.util.TreeSet;
 
-import org.joda.time.LocalDateTime;
+import java.time.LocalDateTime;
```
daschl: are we still supporting pre java 8? or are we on java 8 now completely for spring data

subalakr: we are fully on java 8 now

subalakr:
Merged

PR #141, opened 2017-04-10
DATACOUCH-276 Dont escape the order by property as a whole

subalakr:


src/main/java/org/springframework/data/couchbase/repository/query/support/N1qlUtils.java
```diff
@@ -184,7 +184,16 @@ public static String getDottedPathWithAlternativeFieldNames(PersistentPropertyPa
...
       //FIXME the order property should be converted to its corresponding fieldName
-      Expression orderFieldName = i(orderProperty);
+      String[] orderPropertyParts = orderProperty.split("\\.");
+
+      StringBuilder sb = new StringBuilder();
+      for (String part:orderPropertyParts) {
+        if (sb.length() != 0) {
+          sb.append(".");
+        }
+        sb.append(i(part).toString());
```
daschl: don't think you need the toString() here since the conversion is implicit, but it won't hurt I guess...

subalakr:
Merged

PR #143, opened 2017-04-10
DATACOUCH-147 - Allow to convert Date to ISO-8601 compliant strings

subalakr:
Using the system property
org.springframework.data.couchbase.useISOStringConverterForDate the
conversion can be enabled so this doesn't break backward compatibility

### DIFF
--- a/src/main/asciidoc/entity.adoc
+++ b/src/main/asciidoc/entity.adoc
@@ -271,6 +271,7 @@ A populated object can look like:
 ----
 ====
 
+Optionally, Date can be converted to and from ISO-8601 compliant strings by setting system property `org.springframework.data.couchbase.useISOStringConverterForDate` to true.
 If you want to override a converter or implement your own one, this is also possible. The library implements the general Spring Converter pattern. You can plug in custom converters on bean creation time in your configuration. Here's how you can configure it (in your overridden `AbstractCouchbaseConfiguration`):
 
 .Custom Converters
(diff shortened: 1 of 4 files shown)


src/main/asciidoc/entity.adoc
```diff
@@ -271,6 +271,7 @@ A populated object can look like:
 ----
 ====
 
+Optionally, Date can be converted to and from ISO-8601 complaint strings by setting system property `org.springframework.data.couchbase.useISOStringConverterForDate` to true.
```
simonbasle: typo `complaint` -> `compliant`

src/main/java/org/springframework/data/couchbase/core/convert/DateConverters.java
```diff
@@ -52,9 +56,18 @@ private DateConverters() {
...
-    converters.add(DateToLongConverter.INSTANCE);
+    boolean useISOStringConverterForDate = Boolean.parseBoolean(
+            System.getProperty("org.springframework.data.couchbase.useISOStringConverterForDate", "false"));
+
+    if (useISOStringConverterForDate) {
+      converters.add(DateToStringConverter.INSTANCE);
+      converters.add(StringToDateConverter.INSTANCE);
+    } else {
+      converters.add(DateToLongConverter.INSTANCE);
+      converters.add(NumberToDateConverter.INSTANCE);
```
simonbasle: I wonder if adding the `NumberToDateConverter` still in the case of the property being set to true would work? This would allow to migrate data that was serialized with a `long` encoding for dates to a `String` based format.
("Be liberal in what you accept, and strict in what you produce")

subalakr: You bring up a very good point 👍 If the property is set, reading converter should indeed accept both long and string.

src/main/asciidoc/entity.adoc
subalakr: Thanks, will fix

subalakr:
Merged

PR #144, opened 2017-04-14
DATACOUCH-296 Add support for delete n1ql queries

subalakr:


src/integration/java/org/springframework/data/couchbase/repository/PartyRepository.java
```diff
@@ -50,10 +67,15 @@
       " AND `desc` NOT LIKE '%' || $1 || '%'")
   List<Party> findAllWithPositionalParams(String ex, String inc, long minimumAttendees);
 
+  @Query("#{#n1ql.delete} WHERE #{#n1ql.filter} AND `desc` LIKE '%' || $2 ||  '%' AND attendees < $3" +
+          " AND `desc` NOT LIKE '%' || $1 || '%' #{#n1ql.returning}")
```
daschl: is there also remove with named params?

src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
```diff
@@ -122,10 +122,6 @@ protected static N1qlQuery buildQuery(Statement statement, JsonValue queryPlaceh
   protected Object executeDependingOnType(N1qlQuery query, N1qlQuery countQuery, QueryMethod queryMethod,
       Pageable pageable, Class<?> typeToRead) {
 
-    if (queryMethod.isModifyingQuery()) {
```
daschl: should this check be more focused on checking for delete? i.e. we don't support updates yet, right?

src/main/asciidoc/repository.adoc
```diff
@@ -142,6 +143,8 @@ A few N1QL-specific values are provided through SpEL:
  - `#n1ql.filter` in the WHERE clause adds a criteria matching the entity type with the field that Spring Data uses to store type information.
  - `#n1ql.bucket` will be replaced by the name of the bucket the entity is stored in, escaped in backticks.
  - `#n1ql.fields` will be replaced by the list of fields (eg. for a SELECT clause) necessary to reconstruct the entity.
+ - `#n1ql.delete` will be replaced by the `delete from` statement.
+ - `#n1ql.returning` will be replaced by returning clause needed for reconstructing entity.
```
daschl: hm there probably needs to be more clarity in the docs what and how the return clause affects the returned entity - what are we returning by default? the full thing?

src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
subalakr: I don't think the check is even valid, commons always returns false https://github.com/spring-projects/spring-data-commons/blob/b5af2e9d9685f176d0a5ef0154baeb93b698a556/src/main/java/org/springframework/data/repository/query/QueryMethod.java#L215

src/main/asciidoc/repository.adoc
subalakr: Yes if the user includes returning spel placeholder, it will return the full entity with id and cas

src/integration/java/org/springframework/data/couchbase/repository/PartyRepository.java
subalakr: Yes the params don't really affect the underlying implementation

src/main/java/org/springframework/data/couchbase/repository/query/AbstractN1qlBasedQuery.java
daschl: then maybe get rid of it at all?

daschl:
other than the isModifyingQuery thing looks good to me! 👍

subalakr:
Thanks, I've raised it on commons DATACMNS-1048

subalakr:
isModifyingQuery is used by spring-data-jpa, not by us, commons just provides it as an extension point. I'll create an enhancement ticket to properly handle that.

subalakr:
Merged

PR #153, opened 2017-11-24
DATACOUCH-350 - Generate both blocking & reactive implementations

alexsderkach:
Problem arises, when creating bean definitions for Repository, and both ReactiveCouchbaseRepositoryConfigurationExtension & CouchbaseRepositoryConfigurationExtension are used in Spring Boot autoconfiguration.

First, SimpleReactiveCouchbaseRepository is set as bean definition.
Then, SimpleCouchbaseRepository overrides this definition, despite that Repository uses reactive types.

Thus, repository creation fails, e.g.:
org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [org.springframework.boot.test.autoconfigure.data.couchbase.ExampleDocument] to type [reactor.core.publisher.Mono<?>]


mp911de:
Thanks a lot for your pull request. I'll take it from here.

mp911de:
That's merged, polished and back-ported now.

PR #156, opened 2018-02-14
DATACOUCH-362 Use dynamic proxying

subalakr:
Dynamic proxying for couchbase environment instead of an explicit no shutdown
proxy.


src/main/java/org/springframework/data/couchbase/config/AbstractReactiveCouchbaseConfiguration.java
```diff
@@ -88,11 +89,14 @@ protected CouchbaseConfigurer couchbaseConfigurer() {
     @Override
     @Bean(destroyMethod = "shutdown", name = BeanNames.COUCHBASE_ENV)
     public CouchbaseEnvironment couchbaseEnvironment() {
-        CouchbaseEnvironment env = getEnvironment();
         if (isEnvironmentManagedBySpring()) {
-            return env;
+            return getEnvironment();
+        } else {
```
daschl: three questions: 

- maybe you considered this, but why not just extend the couchbase environment and return false on shutdown?
- is this approach common in spring? never seen it before :)
- how is shutdown handled then if not via the bean destroy method?

subalakr: 1. No the environment instance would be already available and the proxy should just delegate the operations to it.
2. Dynamic proxying- No, I just thought of it.
3. I believe the original reason was the user would pass in the environment preconfigured to getEnvironment and the spring data lifecycle should not shut it down.

PR #162, opened 2018-03-12
DATACOUCH-307 - Support for findDistinctBy.

programmatix:
This is to support syntax like this:

public interface UserRepository extends CouchbaseRepository<User, String> {
  List<AgesOnly> findDistinctBy();

  List<AgesOnly> findDistinctByUsernameContains(String contains);
}

interface AgesOnly {
  int getAge();
}

public class User {
  @Id
  private final String key;

  private final String username;

  private final int age;

  // ... etc
}

The above will find all unique ages.


davidkelly:
Closing for now, per @programmatix.  If anyone asks for this, we can always reopen and go from there.

PR #166, opened 2018-07-12
DATACOUCH-391 - Remove unused imports causing tests build failure

subalakr:
The imported classes have been renamed in commons and they are not
required for these tests.


olivergierke:
That's merged, thanks!

PR #167, opened 2018-07-19
DATACOUCH-393 Add SDK version forward compatibility

bsubhashni:


davidkelly:
@ingenthr - Should we merge this as-is, or is there a logical spot (maybe the `README.adoc`) where we could discuss _how_ to do this?

pivotal-issuemaster:
@bsubhashni Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=167)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=167) to manually synchronize the status of this Pull Request.

daschl:
@davidkelly this is certainly outdated w.r.t versions by now. I think it makes sense to revisit this in a different way: you can probably define a variable in asciidoc which can then be inserted into the docs dynamically? So going forward the variable can be used throughout the docs and only needs to be adjusted in one place?

davidkelly:
I think since we are currently on 2.7.4, and (I believe) will not do a 2.8, maybe this wording is unnecessary.  Also I'd guess that it maybe isn't forward compatible with the coming 3.x.  So, maybe we could say it is fixed at 2.7.4, and compatible with the latest 2.7.x.  But that seems kinda obvious since only the minor version changed.  What do you think @daschl?

davidkelly:
Closing as @daschl agrees with the above.  We will revisit this when 3.0 comes out.

PR #171, opened 2018-08-29
DATACOUCH-401 Incorrect return type for reactive string based N1ql co…

subalakr:
…unt query

Motivation
----------
Reactive string based N1QL count query incorrectly returns the map after
the single result projection.

Changes
-------
Map the observable returning from the N1ql query result deserialized to
a map by converting to the correct return type using the conversion
service. Verified by integration tests.

### DIFF
--- a/src/integration/java/org/springframework/data/couchbase/repository/ReactiveN1qlCouchbaseRepositoryTests.java
+++ b/src/integration/java/org/springframework/data/couchbase/repository/ReactiveN1qlCouchbaseRepositoryTests.java
@@ -111,4 +111,16 @@ public class ReactiveN1qlCouchbaseRepositoryTests {
 		}
 		assertNotNull("Expected to find several parties", previousDesc);
 	}
+
+    @Test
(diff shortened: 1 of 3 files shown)


DanielFran:
@mp911de Is this commit duplicated with https://github.com/spring-projects/spring-data-couchbase/commit/5af4dae334d350d99858b2ff2a155fb1e53b0cbe

I do not see reference for DATACOUCH-401 in changelog?

mp911de:
@DanielFran This is because the ticket state of DATACOUCH-401 is still in progress and it's not assigned to a version. Spring Data Couchbase is maintained by Couchbase, maybe @subalakr can help here.

subalakr:
Yes it has been fixed in lovelace GA. Apologies for not updating the ticket

DanielFran:
@subalakr Thanks. So this PR should be deleted to avoid confusions?

subalakr:
Yes closing it.

PR #181, opened 2019-03-13
Remove stray import IsNewStrategyFactory

subalakr:
This has been removed from commons and causing build failure


odrotbohm:
Lovely, @subalakr. I was just about to look into this one :)

PR #198, opened 2019-06-04
#DATACOUCH-206 set scan consistency on repository method

johannesjasper:
When using N1QL with spring data repositories, there is no way to set the scan consistency individually, only a global default could be set.

This adds a `@WithConsistency` annotation that allows you to set the scan consistency per query.
When building the query parameters in `AbstractN1qlBasedQuery` and `ReactiveAbstractN1qlBasedQuery` this annotation is evaluated, the current behavior (using the global default) is used as fallback in case no annotation was used.

Currently the added methods are duplicated in `AbstractN1qlBasedQuery` and `ReactiveAbstractN1qlBasedQuery` since they don't share a common interface. The same is true for `CouchbaseOperations` and `RxJavaCouchbaseOperations`. I could extract a common interface from each and provide a default implementation.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.couchbase.core.query;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.data.annotation.QueryAnnotation;
+
+import com.couchbase.client.java.query.consistency.ScanConsistency;
+
+/**
+ * Annotation to set the scan consistency of N1QL queries with Couchbase.
+ * This controls whether couchbase waits for all changes to be processed by an index
+ * or whether stale results are acceptable.
+ * <p/>
+ * If not set, the {@link Consistency#DEFAULT_CONSISTENCY default consistency} is used.
+ *
+ * @author Johannes Jasper.
+ */
+@Documented
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@QueryAnnotation
+public @interface WithConsistency {
+
+  ScanConsistency value() default ScanConsistency.REQUEST_PLUS;
+
+}
--- a/src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
@@ -16,22 +16,21 @@
 
 package org.springframework.data.couchbase.core.query;
 
+import com.couchbase.client.java.query.consistency.ScanConsistency;
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-
 import org.springframework.data.annotation.QueryAnnotation;
-
-import com.couchbase.client.java.query.consistency.ScanConsistency;
+import org.springframework.data.couchbase.config.AbstractCouchbaseConfiguration;
 
 /**
  * Annotation to set the scan consistency of N1QL queries with Couchbase.
  * This controls whether couchbase waits for all changes to be processed by an index
  * or whether stale results are acceptable.
  * <p/>
- * If not set, the {@link Consistency#DEFAULT_CONSISTENCY default consistency} is used.
+ * If not set, the default consistency set in {@link AbstractCouchbaseConfiguration#getDefaultConsistency()} is used.
  *
  * @author Johannes Jasper.
  */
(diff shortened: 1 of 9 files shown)


dnault:
Thank you @johannesjasper, this is good stuff. One question:

Is it ever useful to apply the `@WithConsistency` annotation to a method that is not also annotated with `@Query`? If not, then perhaps it would be cleaner to add a "consistency" attribute to `@Query` instead of using a separate annotation. That would avoid the possibility of a user applying `@WithConsistency` to an inappropriate method.

johannesjasper:
Cheers! Thank you for your feedback.

Even if your queries are being derived from the method name (e.g. `findByName`) I find it useful to specify the scan consistency. The problem I wanted to tackle with this PR is that so far you can only set a single global consistency.
By implementing this feature on the `AbstractN1qlBasedQuery` the `@WithConsistency` annotation works both on String based queries, as well as queries derived from method name.

On a technical note, I tried to add a method `consistency()` to `@Query` but I was not able to set a meaningful default. 
Omitting the default would mean breaking the API which, I guess, is unacceptable.
I could set it to 
```java
ScanConsistency consistency() default ScanConsistency.STATEMENT_PLUS;
```
I can **not**, however, set it to 
```java
Consistency consistency() default Consistency.DEFAULT_CONSISTENCY.n1qlConsistency();
```

Either way, this would not reflect the default consistency set in the couchbase config.
What I wanted to avoid is a situation in which a user changed the default consistency (let's say to `NOT_BOUNDED`) and then uses the `@Query` annotation. In that case I would not known which consistency to pick, the consistency from the `@Query` annotation (which cannot be null) or the default consistency.

If you have any suggestion on how to implement this, I'd be more than happy to integrate it!

dnault:
> So yes, I think this annotation is useful independent of whether you use the `@Query` annotation or not.

I see now. Thank you.

> On a technical note, I tried to add a method `consistency()` to `@Query` but I was not able to find a meaningful default.

One way to do this would be to define a separate Consistency enum just for the annotation attribute, where one of the values is  a DEFAULT (null). A value of DEFAULT can be replaced with the global default at runtime. Something like:

```
public @interface Query {
  ...
  Consistency consistency() default Consistency.DEFAULT;

  enum Consistency {
    NOT_BOUNDED,
    REQUEST_PLUS,
    STATEMENT_PLUS,
    DEFAULT 
  }
}
```

I'm not necessarily recommending that approach in this case; just offering it as a possible solution to the technical challenge you described.

src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.data.annotation.QueryAnnotation;
+
+import com.couchbase.client.java.query.consistency.ScanConsistency;
+
+/**
+ * Annotation to set the scan consistency of N1QL queries with Couchbase.
+ * This controls whether couchbase waits for all changes to be processed by an index
+ * or whether stale results are acceptable.
+ * <p/>
+ * If not set, the {@link Consistency#DEFAULT_CONSISTENCY default consistency} is used.
```
dnault: Doesn't the actual value used come from `getCouchbaseOperations().getDefaultConsistency()` which may be different than `Consistency.DEFAULT_CONSISTENCY` (`READ_YOUR_OWN_WRITES` / `STATEMENT_PLUS`)?

src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @author Johannes Jasper.
+ */
+@Documented
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@QueryAnnotation
+public @interface WithConsistency {
+
+  ScanConsistency value() default ScanConsistency.REQUEST_PLUS;
```
dnault: It's probably worth mentioning this default value in the Javadoc.

dnault:
Just to clarify, I'm not requesting that you add the `consistency()` attribute to `@Query`. I think it's fine for `@WithConsistency` to be the only way to specify the per-method consistency.

src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
johannesjasper: Good point, I pointed the javadoc to `AbstractCouchbaseConfiguration#getDefaultConsistency()`

src/main/java/org/springframework/data/couchbase/core/query/WithConsistency.java
johannesjasper: Instead of documenting the chosen default I simply omitted it: now when using the `@WithConsistency` annotation, you have to specify one.

johannesjasper:
I just realized that I did not update the documentation. Should I do that?

dnault:
> I just realized that I did not update the documentation. Should I do that?

Yes, please. I think this is the relevant section:

https://github.com/spring-projects/spring-data-couchbase/blob/master/src/main/asciidoc/repository.adoc#querying-with-consistency

dnault:
@johannesjasper Just wanted to let you know that if I don't respond next week it's because I will be on vacation away from a computer.

johannesjasper:
I added some documentation with an example that shows how `@WithConsistency`, `@Query`, and query derivation work together.
I also left a note that this works for N1QL queries only, for now. I'd like to tackle that in a different PR if that is fine with you.

mp911de:
On a related note: We have a similar annotation in Spring Data Cassandra with [`@Consistency`](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/Consistency.java) applied for query derivation (part-tree query) and annotated `@Query` methods.

Attributes that are valid only on the scope of string queries are typically part of `@Query`. In some cases (e.g. [`@ExistsQuery`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/ExistsQuery.java)), we provide annotation aliases.

johannesjasper:
> We have a similar annotation in Spring Data Cassandra with [`@Consistency`](https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/Consistency.java) applied for query derivation (part-tree query) and annotated `@Query` methods.
Yes that was part of my inspiration :)

> Attributes that are valid only on the scope of string queries are typically part of `@Query`
Maybe I misunderstood you, but the `@WithConsistency` annotation is valid for String based queries (using `@Query`) as well as derived (part-tree) queries

johannesjasper:
Great, thanks for the approval @dnault . Should somebody else review this or can it be merged?

PR #163, opened 2018-04-19
DATACOUCH-356 - Fix SimpleReactiveCouchbaseRepository.deleteAll()

sixdouglas:
- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACOUCH).
- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [ ] You submit test cases (unit or integration tests) that back your changes.
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

See also: https://jira.spring.io/browse/DATACOUCH-356

PR #154, opened 2017-12-04
DATACOUCH-356 - Retrieve AsyncViewRow from Observable

rmsantana:
Retrieve AsyncViewRow from Observable instead of using cast

- [x] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACOUCH).
- [x] You use the code formatters provided [here](https://github.com/spring-projects/spring-data-build/tree/master/etc/ide) and have them applied to your changes. Don’t submit any formatting related changes.
- [x] You submit test cases (unit or integration tests) that back your changes.
- [x] You added yourself as author in the headers of the classes you touched. Amend the date range in the Apache license header if needed. For new types, add the license header (copy from another file and set the current year only).

PR #200, opened 2019-06-26
DATACOUCH-356 Fix for cast exception in deleteAll

davidkelly:
Added a simple test for it as well.  After doing this, I noticed #163 is basically the same fix (this was a redo of #154).  So, I will close #163 and add Douglas Six to the authors list.


davidkelly:
ugh - got my branch in odd state - new one coming...

PR #201, opened 2019-06-26
DATACOUCH-356 Fix cast issue in deleteAll

davidkelly:
Plus a test.  I didn't notice that Douglas Six did basically the
same thing in another PR (this was a re-do of an earlier PR - #154 which
didn't quite do what we'd like).  So since I already had it done
and had a test, I closed that PR (#163) and added him to the author
list.

Also - repository.count() raises a NoSuchElementException after you
delete all the documents.  That seems like a bug, but we will deal
with that separately.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/repository/SimpleReactiveCouchbaseRepositoryDeleteAllIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/repository/SimpleReactiveCouchbaseRepositoryDeleteAllIntegrationTests.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


dnault:
The change itself looks good! There's a redundant commit for DATACOUCH-206 in this PR... might need to rebase your feature branch onto the latest master and update this PR.

davidkelly:
Thx David -- I missed that -- I'll see what I can do

davidkelly:
ohhh that happened after I made the PR I think -- interesting

davidkelly:
ok that did it

src/test/java/org/springframework/data/couchbase/repository/SimpleReactiveCouchbaseRepositoryDeleteAllIntegrationTests.java
```diff
@@ -0,0 +1,94 @@
...
+        repository.save(instance).block();
+
+        // we put a doc in, lets be sure the count reflects that there is
+        // at least one doc in there...
+        assertTrue(getCount() > 0L);
+
+        repository.deleteAll().block();
+
+        try {
+            getCount();
```
daschl: views are eventually consistent, so this might race at the first view attempts.. I think it's good for now, but keep in mind should you see it failing occasionally this might be the reason.

daschl:
oh, looks like I am too late to the party hehe

mp911de:
Please mind the commit message format. That should be ` DATACOUCH-356 - Fix cast issue in deleteAll.` instead of ` DATACOUCH-356 Fix cast issue in deleteAll (#201)`.

src/test/java/org/springframework/data/couchbase/repository/SimpleReactiveCouchbaseRepositoryDeleteAllIntegrationTests.java
davidkelly: ohh - great point.  Thx -- maybe I need to rethink this test a bit, when I update it to mixup the count issue...

PR #202, opened 2019-06-27
DATACOUCH-462 ReactiveCouchbaseRepository count() will raise exceptio…

davidkelly:
…n when bucket is empty

Small cleanup in the deleteAll test (which actually tests this as well), while
we are there.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
@@ -21,7 +21,6 @@ import java.io.Serializable;
 import com.couchbase.client.java.document.json.JsonArray;
 import com.couchbase.client.java.error.DocumentDoesNotExistException;
 import com.couchbase.client.java.view.AsyncViewResult;
-import com.couchbase.client.java.view.AsyncViewRow;
 import com.couchbase.client.java.view.ViewQuery;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
```diff
@@ -238,7 +238,8 @@ protected Flux mapFlux(Observable observable) {
                 .queryView(query)
                 .flatMap(AsyncViewResult::rows)
                 .map(asyncViewRow ->
-                        Long.valueOf(asyncViewRow.value().toString())).toSingle());
+                        Long.valueOf(asyncViewRow.value().toString()))
+                .switchIfEmpty(Observable.just(Long.valueOf(0L))).toSingle());
```
dnault: NIT: Instead of `Observable.just(Long.valueOf(0L))` you can take advantage of auto-boxing and write `Observable.just(0L)`.

davidkelly: ohhh nice - I see that now.  Will update now!

mp911de:
Note: 1st line of commit messages also ends with a period, see also https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc#commit-messages.

davidkelly:
@daschl and @mp911de - thx I'll amend the commit!

PR #206, opened 2019-09-25
DATACOUCH-479 - Use version property for testcontainers in pom.xml.

davidkelly:


davidkelly:
Thx @mp911de - nice catch!  Done.

PR #207, opened 2019-10-31
DATACOUCH-479 - Fix issue with testcontainers usage.

davidkelly:
After updating the POM to not have a fixed testcontainers version, we
discovered the tests were failing with later versions of testcontainers.  It
seems that they updated how they did WaitStrategy, deprecating the old way, so
our WaitStrategy was not being called.  That means no setup was done to
the cluster (passwords, buckets, etc...).  Simple fix was to move to the
new interface.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/CouchbaseWaitStrategy.java
+++ b/src/test/java/org/springframework/data/couchbase/CouchbaseWaitStrategy.java
@@ -7,11 +7,11 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
 import com.couchbase.client.java.util.features.Version;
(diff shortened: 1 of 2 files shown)


mp911de:
In general, would it make sense to switch from `FixedHostPortGenericContainer ` to the `CouchbaseContainer` implementation?

daschl:
@mp911de in theory yes, although the testcontainer project reports intermittent test failures with a wait strategy which we didn't get to the bottom of yet. So not sure if in the short term this would cause more problems than it solves? In the long run I think so

mp911de:
Thanks for the background. My question was more out of curiosity than a necessity.

davidkelly:
Actually I'll probably try to fix the CouchbaseContainer implementation soon, then we can maybe use it here...

PR #208, opened 2019-11-11
DATACOUCH-484 - Thread safety issue using findBy.

davidkelly:
Support for positional parameters necessitated caching them, however
that was not done in a threadsafe fashion.  ThreadLocal storage is
sufficient to take care of the issue.  Added test for it while at it.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/query/PartTreeN1qlBasedQuery.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/PartTreeN1qlBasedQuery.java
@@ -50,16 +50,21 @@ import org.springframework.util.Assert;
 public class PartTreeN1qlBasedQuery extends AbstractN1qlBasedQuery {
 
   private final PartTree partTree;
-  private JsonValue placeHolderValues;
+  private ThreadLocal<JsonValue> placeHolderValues;
 
   public PartTreeN1qlBasedQuery(CouchbaseQueryMethod queryMethod, CouchbaseOperations couchbaseOperations) {
     super(queryMethod, couchbaseOperations);
     this.partTree = new PartTree(queryMethod.getName(), queryMethod.getEntityInformation().getJavaType());
+    this.placeHolderValues = new ThreadLocal<JsonValue>() {
+      @Override public JsonValue initialValue() {
+        return JsonArray.create();
+      }
+    };
   }
 
   @Override
   protected JsonValue getPlaceholderValues(ParameterAccessor accessor) {
-    return this.placeHolderValues;
+    return this.placeHolderValues.get();
   }
 
   @Override
@@ -70,7 +75,7 @@ public class PartTreeN1qlBasedQuery extends AbstractN1qlBasedQuery {
     N1qlCountQueryCreator queryCountCreator = new N1qlCountQueryCreator(partTree, accessor, countFrom,
         getCouchbaseOperations().getConverter(), getQueryMethod());
     Statement statement = queryCountCreator.createQuery();
-    this.placeHolderValues = queryCountCreator.getPlaceHolderValues();
+    this.placeHolderValues.set(queryCountCreator.getPlaceHolderValues());
     return statement;
   }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/couchbase/repository/query/PartTreeN1qlBasedQuery.java
```diff
@@ -50,16 +50,21 @@
...
   private final PartTree partTree;
-  private JsonValue placeHolderValues;
+  private ThreadLocal<JsonValue> placeHolderValues;
 
   public PartTreeN1qlBasedQuery(CouchbaseQueryMethod queryMethod, CouchbaseOperations couchbaseOperations) {
     super(queryMethod, couchbaseOperations);
     this.partTree = new PartTree(queryMethod.getName(), queryMethod.getEntityInformation().getJavaType());
+    this.placeHolderValues = new ThreadLocal<JsonValue>() {
+      @Override public JsonValue initialValue() {
+        return JsonArray.create();
```
daschl: looks like this was previously uninitialized (the inner value) should this still be null or does this any harm? My assumption is no but still checking

src/test/java/org/springframework/data/couchbase/repository/N1qlCouchbaseRepositoryIntegrationTests.java
```diff
@@ -90,6 +97,41 @@ public void cleanUp() {
...
+  public void doShouldBeThreadsafe() {
+    int threads = 50;
+    ExecutorService service = Executors.newFixedThreadPool(threads);
+    List<Callable<Boolean>> callables = new ArrayList<>();
+    for (int thread = 0; thread < threads; ++thread) {
+      final int counter = thread;
+      Callable<Boolean> booleanSupplier = () -> {
+        String expectedName = "party like it's 199" + counter%12;
+        String foundName = partyRepository.findByName(expectedName).get(0).getName();
+        System.out.println("expected = " + expectedName  + " found = " + foundName);
```
daschl: make a real assertEquals and not println so it fails the test if needed and not spam the logs

davidkelly: I can just remove this line -- I'd popped it in to make understanding what happened easier, during development.

davidkelly:
@mp911de - The design that was there cached the parameters rather than keeping them on the stack as it passed through the various calls.  I felt like just making this cached value thread-local was safer than re-working the entire set of calls (that span quite a few objects).  

Part of my goal was getting a fix to a customer, so this seemed safer.  But certainly it isn't the optimal design.  Had I more time I'd rework the N1QL utility classes.  I suppose I can still do so, instead of this, but not right away.

mp911de:
Thanks a lot for the details. Let's please address that issue for the 4.0 line.

davidkelly:
@mp911de - how about we put this in master and get it out to the customer.  Then, in the new SDK3 version which is coming soon, we will do this right and keep the parameters on the call stack.  LMK what you think.

src/main/java/org/springframework/data/couchbase/repository/query/PartTreeN1qlBasedQuery.java
davidkelly: This does no harm - IIRC not initializing it was a problem.

mp911de:
Works for me @davidkelly, that's what I meant with let's fix it for the 4.0 (SDK3) line.

davidkelly:
@mp911de Thx!  I think my browser hadn't noticed your comment before I made the second one 👍

mikereiche:
This issue is still in 3.0.12-RELEASE through 3.2.2-RELEASE

mp911de:
Please note that only the 3.1 and 3.2 lines (Lovelace and Moore) are maintained and actively released.

mikereiche:
Thanks.  I checked and found the 413 issue in 3.1.0, 3.1.1 and 3.1.2 only.  I'll look at what change was made for 3.1.3.

PR #210, opened 2019-12-26
DATACOUCH-496 - Support JSR303 Validation in ReactiveCouchbaseRepository

l30thelion:
DATACOUCH-496 - Support JSR303 Validation in ReactiveCouchbaseRepository

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/RxJavaCouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/RxJavaCouchbaseTemplate.java
@@ -52,6 +52,7 @@ import rx.functions.Func3;
  * @author Subhashni Balakrishnan
  * @author Mark Paluch
  * @author Alex Derkach
+ * @author Leo Duarte
  * @since 3.0
  */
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/couchbase/core/RxJavaCouchbaseTemplate.java
```diff
@@ -66,7 +68,8 @@
     private final TranslationService translationService;
     private Consistency configuredConsistency = Consistency.DEFAULT_CONSISTENCY;
     private WriteResultChecking writeResultChecking = DEFAULT_WRITE_RESULT_CHECKING;
-
+    private Validator validator;
```
mp911de: Introducing `Validator` at Template API level requires the Validation API to be present on the class path of each project that wants to use Spring Data Couchbase. We cannot do that.

Instead, the validation should live in an optional component such as a validation event listener. Take a look at [Mongo's `ValidatingMongoEventListener`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ValidatingMongoEventListener.java).

PR #214, opened 2020-04-13
DATACOUCH-519 - reinstate auditing to template

mikereiche:
This reinstates the auditing that was present before the Reactive refactoring.

- [x] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACOUCH). https://jira.spring.io/browse/DATACOUCH-519

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
@@ -17,9 +17,7 @@
 package org.springframework.data.couchbase.core;
 
 import org.springframework.beans.BeansException;
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationContextAware;
-import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.*;
 import org.springframework.dao.DataAccessException;
 import org.springframework.dao.support.PersistenceExceptionTranslator;
 import org.springframework.data.couchbase.CouchbaseClientFactory;
@@ -33,6 +31,14 @@ import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProper
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.lang.Nullable;
 
+/**
+ * Implements Couchbase operations
+ * findBy, insertBy, upsertBy, replaceBy, removeBy, existsBy
+ *
+ * @author Michael Nitschinger
+ * @author Michael Reiche
+ * @since 3.0
+ */
 public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContextAware {
 
 	private final CouchbaseClientFactory clientFactory;
@@ -162,6 +168,8 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		prepareIndexCreator(applicationContext);
+		templateSupport.setApplicationContext(applicationContext);
+		reactiveCouchbaseTemplate.setApplicationContext(applicationContext);
 	}
 
 	private void prepareIndexCreator(ApplicationContext context) {
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
@@ -47,8 +47,8 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 	private final CouchbaseTemplateSupport templateSupport;
 	private final MappingContext<? extends CouchbasePersistentEntity<?>, CouchbasePersistentProperty> mappingContext;
 	private final ReactiveCouchbaseTemplate reactiveCouchbaseTemplate;
-	private @Nullable CouchbasePersistentEntityIndexCreator indexCreator;
-
+	private @Nullable
+	CouchbasePersistentEntityIndexCreator indexCreator;
 
 	public CouchbaseTemplate(final CouchbaseClientFactory clientFactory, final CouchbaseConverter converter) {
 		this.clientFactory = clientFactory;
@@ -176,7 +176,8 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 		String[] indexCreators = context.getBeanNamesForType(CouchbasePersistentEntityIndexCreator.class);
 
 		for (String creator : indexCreators) {
-			CouchbasePersistentEntityIndexCreator creatorBean = context.getBean(creator, CouchbasePersistentEntityIndexCreator.class);
+			CouchbasePersistentEntityIndexCreator creatorBean = context.getBean(creator,
+					CouchbasePersistentEntityIndexCreator.class);
 			if (creatorBean.isIndexCreatorFor(mappingContext)) {
 				return;
 			}
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
@@ -16,23 +16,41 @@
 
 package org.springframework.data.couchbase.core;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
 import org.springframework.data.couchbase.core.convert.CouchbaseConverter;
 import org.springframework.data.couchbase.core.convert.translation.JacksonTranslationService;
 import org.springframework.data.couchbase.core.convert.translation.TranslationService;
 import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
+import org.springframework.data.couchbase.core.mapping.event.*;
 import org.springframework.data.couchbase.repository.support.MappingCouchbaseEntityInformation;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
+import org.springframework.data.mapping.callback.EntityCallbacks;
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.data.mapping.model.ConvertingPropertyAccessor;
+import org.springframework.util.Assert;
 
-public class CouchbaseTemplateSupport {
+/**
+ * encode/decode support for CouchbaseTemplate
+ *
+ * @author Michael Nitschinger
+ * @author Michael Reiche
+ * @since 3.0
+ */
+public class CouchbaseTemplateSupport implements ApplicationContextAware {
+	private static final Logger LOG = LoggerFactory.getLogger(CouchbaseTemplateSupport.class);
 
 	private final CouchbaseConverter converter;
 	private final MappingContext<? extends CouchbasePersistentEntity<?>, CouchbasePersistentProperty> mappingContext;
 	// TODO: this should be replaced I think
 	private final TranslationService translationService;
+	private ApplicationContext applicationContext;
+	EntityCallbacks entityCallbacks;
 
 	public CouchbaseTemplateSupport(final CouchbaseConverter converter) {
 		this.converter = converter;
@@ -41,8 +59,12 @@ public class CouchbaseTemplateSupport {
 	}
 
 	public CouchbaseDocument encodeEntity(final Object entityToEncode) {
+		maybeEmitEvent(new BeforeConvertEvent<>(entityToEncode));
+		Object maybeNewEntity = maybeCallBeforeConvert(entityToEncode, "");
 		final CouchbaseDocument converted = new CouchbaseDocument();
-		converter.write(entityToEncode, converted);
+		converter.write(maybeNewEntity, converted);
+		maybeCallAfterConvert(entityToEncode, converted, "");
+		maybeEmitEvent(new BeforeSaveEvent<>(entityToEncode, converted));
 		return converted;
 	}
 
@@ -52,7 +74,8 @@ public class CouchbaseTemplateSupport {
 
 		T readEntity = converter.read(entityClass, (CouchbaseDocument) translationService.decode(source, converted));
 		final ConvertingPropertyAccessor<T> accessor = getPropertyAccessor(readEntity);
-		CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(readEntity.getClass());
+		CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(
+				readEntity.getClass());
 
 		if (persistentEntity.getVersionProperty() != null) {
 			accessor.setProperty(persistentEntity.getVersionProperty(), cas);
@@ -62,7 +85,8 @@ public class CouchbaseTemplateSupport {
 
 	public void applyUpdatedCas(final Object entity, final long cas) {
 		final ConvertingPropertyAccessor<Object> accessor = getPropertyAccessor(entity);
-		final CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(entity.getClass());
+		final CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(
+				entity.getClass());
 		final CouchbasePersistentProperty versionProperty = persistentEntity.getVersionProperty();
 
 		if (versionProperty != null) {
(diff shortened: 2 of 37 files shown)


src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -33,22 +31,29 @@
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.lang.Nullable;
 
-public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContextAware {
+/*
+ * @author Michael Nitschinger
+ * @author Michael Reiche
+ */
+
+public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContextAware /*, ApplicationEventPublisherAware */ {
```
daschl, line 39: why is this commented out? please remove if leftover

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -33,22 +31,29 @@
...
+	private @Nullable
+	CouchbasePersistentEntityIndexCreator indexCreator;
+	//private ApplicationEventPublisher eventPublisher;// TODO: this field not needed?
 
 	public CouchbaseTemplate(final CouchbaseClientFactory clientFactory, final CouchbaseConverter converter) {
 		this.clientFactory = clientFactory;
 		this.converter = converter;
 		this.exceptionTranslator = clientFactory.getExceptionTranslator();
 		this.templateSupport = new CouchbaseTemplateSupport(converter);
+		// TODO: why doesn't the ReactiveCouchbaseTemplate use the same templateSupport ?
```
daschl, line 56: still need this?

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -33,22 +31,29 @@
...
 	private final CouchbaseConverter converter;
 	private final PersistenceExceptionTranslator exceptionTranslator;
 	private final CouchbaseTemplateSupport templateSupport;
 	private final MappingContext<? extends CouchbasePersistentEntity<?>, CouchbasePersistentProperty> mappingContext;
 	private final ReactiveCouchbaseTemplate reactiveCouchbaseTemplate;
-	private @Nullable CouchbasePersistentEntityIndexCreator indexCreator;
-
+	private @Nullable
+	CouchbasePersistentEntityIndexCreator indexCreator;
+	//private ApplicationEventPublisher eventPublisher;// TODO: this field not needed?
```
daschl, line 49: leftover?

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -178,4 +185,12 @@ private void prepareIndexCreator(ApplicationContext context) {
 			((ConfigurableApplicationContext) context).addApplicationListener(indexCreator);
 		}
 	}
+/*
```
daschl, line 188: leftover?

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -16,23 +16,40 @@
...
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.data.mapping.model.ConvertingPropertyAccessor;
+import org.springframework.util.Assert;
 
-public class CouchbaseTemplateSupport {
+/*
+ * @author Michael Nitschinger
+ * @author Michael Reiche
+ */
+
```
daschl, line 42: please add a proper docblock and no whitespace

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -82,4 +103,74 @@ public String getJavaNameForEntity(final Class<?> clazz) {
...
+	 * @param entityCallbacks must not be {@literal null}.
+	 * @throws IllegalArgumentException if the given instance is {@literal null}.
+	 * @since 2.2
+	 */
+	public void setEntityCallbacks(EntityCallbacks entityCallbacks) {
+		Assert.notNull(entityCallbacks, "EntityCallbacks must not be null!");
+		this.entityCallbacks = entityCallbacks;
+	}
+
+	void maybeEmitEvent(CouchbaseMappingEvent<?> event) {
```
daschl, line 136: formatting is way off here

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperation.java
```diff
@@ -17,15 +17,20 @@
...
+ * @author Michael Reiche
+ */
+
 public interface ExecutableFindByIdOperation {
 
 	<T> ExecutableFindById<T> findById(Class<T> domainType);
 
 	interface TerminatingFindById<T> {
 
-		T one(String id);
```
daschl, line 26: this is a separate change, right? Why do we need convert from string to object on ids in the auditing change?

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
```diff
@@ -79,6 +84,7 @@ public ReactiveFindByQueryOperationSupport(final ReactiveCouchbaseTemplate templ
 		public Flux<T> all() {
 			return Flux.defer(() -> {
 				String statement = assembleEntityQuery(false);
+				System.out.println(statement);
```
daschl, line 87: leftovers

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: It's not needed, but I still have that question. Why does doesn't  ReactiveCouchbaseTemplate use the same templateSupport as CouchbaseTemplate?

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
mikereiche: I'm using https://github.com/krasa/EclipseCodeFormatter#instructions. Maybe I have the wrong settings for it?

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperation.java
mikereiche: My test case included a PersonRepository<Person,UUID> which will not work.  one(<T> object) methods fail when @Id is anything other than String, because the code called relies on id being a String.  Without refactoring all the classes and interfaces to be parameterized with <T,ID> instead of just <T>, simply changing String to Object and (eventually) calling id.toString() 

Actually a better solution would be to make the Entity objects parameterized by ID so the repository would be

PersonRepository<Person<UUID>>

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
mikereiche: Oops.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: It can come out. I was confused that something that needs to publish application events does not need to be "ApplicationEventPublisherAware".  I suspect that this currently works almost by accident, as CouchbaseTemplate (via CouchbaseTemplateSupport) publishes events using applicationContext.publishEvent(), and not via eventPublisher.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -162,13 +168,16 @@ RuntimeException potentiallyConvertRuntimeException(RuntimeException ex) {
 	@Override
 	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
 		prepareIndexCreator(applicationContext);
+		templateSupport.setApplicationContext(applicationContext);
```
daschl, line 171: Why do we need to set the application context und support and reactive like this, can't we make both application context aware itself and let it be injected by spring?

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -16,23 +16,41 @@
...
-public class CouchbaseTemplateSupport {
+/**
+ * encode/decode support for CouchbaseTemplate
+ *
+ * @author Michael Nitschinger
+ * @author Michael Reiche
+ * @since 3.0
+ */
+public class CouchbaseTemplateSupport implements ApplicationContextAware {
+	private static final Logger LOG = LoggerFactory.getLogger(CouchbaseTemplate.class);
```
daschl, line 46: wrong logger class name

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -162,13 +168,16 @@ RuntimeException potentiallyConvertRuntimeException(RuntimeException ex) {
...
+		reactiveCouchbaseTemplate.setApplicationContext(applicationContext);
```
daschl, line 172: I think the reactive template already gets called from spring

src/main/java/org/springframework/data/couchbase/repository/auditing/CouchbaseAuditingRegistrar.java
```diff
@@ -0,0 +1,100 @@
...
+import org.springframework.core.type.AnnotationMetadata;
+import org.springframework.data.auditing.AuditingHandler;
+import org.springframework.data.auditing.IsNewAwareAuditingHandler;
+import org.springframework.data.auditing.config.AuditingBeanDefinitionRegistrarSupport;
+import org.springframework.data.auditing.config.AuditingConfiguration;
+import org.springframework.data.config.ParsingUtils;
+import org.springframework.data.couchbase.config.BeanNames;
+import org.springframework.data.couchbase.core.mapping.CouchbaseMappingContext;
+import org.springframework.data.couchbase.core.mapping.event.AuditingEventListener;
+//import org.springframework.data.support.IsNewStrategyFactory;
```
daschl, line 34: remove?

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateKeyValueIntegrationTests.java
```diff
@@ -25,17 +26,30 @@
...
 
+/**
+ * KV tests
+ *
+ * Theses tests rely on a cb server running.
+ *
+ * @author Michael Nitschinger
+ * @author Michael Reiche
+ */
+@IgnoreWhen(clusterTypes = ClusterType.MOCKED)
```
daschl, line 52: why are we  ignored all the kv tests for the mock? most of them should work, only exists doesn't iirc.

src/test/java/org/springframework/data/couchbase/domain/AbstractEntity.java
```diff
@@ -0,0 +1,75 @@
+package org.springframework.data.couchbase.domain;
```
daschl, line 1: license header missing

src/test/java/org/springframework/data/couchbase/domain/AbstractEntity.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.UUID;
+
+/**
+ * @author Oliver Gierke
+ *
+ */
+//@QuerySupertype
+@Document
+//@Data
```
daschl, line 19: some cleanup?

src/test/java/org/springframework/data/couchbase/domain/Config.java
```diff
@@ -0,0 +1,49 @@
+package org.springframework.data.couchbase.domain;
```
daschl, line 1: license header missing (maybe double check all new files)

src/test/java/org/springframework/data/couchbase/domain/NaiveAuditorAware.java
```diff
@@ -0,0 +1,23 @@
+package org.springframework.data.couchbase.domain;
+
+import org.springframework.data.domain.AuditorAware;
+
+import java.util.Optional;
+
+//import org.springframework.security.core.Authentication;
```
daschl, line 7: unused cleanup and license header

src/test/java/org/springframework/data/couchbase/domain/Person.java
```diff
@@ -0,0 +1,95 @@
...
+        if(lastModifiedBy != null) sb.append(", lastModifiedBy : "+lastModifiedBy);
+        if(lastModification != 0) sb.append(", lastModification : "+lastModification);
+        sb.append("}");
+        return sb.toString();
+    }
+
+    static String optional(String name, Optional<String> obj){
+        if(obj != null)
+            if(obj.isPresent())
+                return("  "+name+ ": '"+obj.get()+"'\n");
```
daschl, line 90: formatting off (can you please rerun the formatter on all files when done?)

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: removed

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperation.java
mikereiche: removed this change.  It's now in a separate branch.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: removed.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: removed

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: CouchbaseTemplateSupport isn't a bean, it isn't create or managed by Spring.

CouchbaseTemplate (and ReactiveCouchbaseTemplate) are beans. They are created by Spring, and since they are ApplicationContextAware, Spring calls their setApplicationContext() method after creating them.

During the construction of CouchbaseTemplate (and ReactiveCouchbaseTemplate), the CouchbaseTemplateSupport object is created like this :

	this.templateSupport = new CouchbaseTemplateSupport(converter);

CouchbaseTemplate (and Reactive) do not have the application context at that point. They only get the applicationcontext afterwards, when their setApplicationContext() method is called.  And that's a good place to propagate the application context to the CouchbaseTemplateSupport object.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
mikereiche: I used CouchbaseTemplate as it does most of the work for CouchbaseTemplate, and if someone wanted to trace/log/debug CouchbaseTemplate, they would certainly want to see what was going on in CouchbaseTemplateSupport.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: reactiveCouchbaseTemplate is created and managed by CouchbaseTemplate. 

this.reactiveCouchbaseTemplate = new ReactiveCouchbaseTemplate(clientFactory, converter);

src/main/java/org/springframework/data/couchbase/repository/auditing/CouchbaseAuditingRegistrar.java
mikereiche: Removed.

src/test/java/org/springframework/data/couchbase/domain/Config.java
mikereiche: ok.

src/test/java/org/springframework/data/couchbase/domain/Person.java
mikereiche: ok

src/test/java/org/springframework/data/couchbase/domain/AbstractEntity.java
mikereiche: Left over from some query dsl testing. Removed.

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateKeyValueIntegrationTests.java
mikereiche: I went back and fixed this.  The connectionString etc. in the Config class was not leveraging the connectionString etc from ClusterAwareIntegrationTests.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: removed.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
mikereiche: fixed all the docblocks

src/test/java/org/springframework/data/couchbase/domain/AbstractEntity.java
mikereiche: I've added the license header to all the files that were missing it.

src/test/java/org/springframework/data/couchbase/domain/NaiveAuditorAware.java
mikereiche: Those commented imports are to show the classes that would be used when making a proper class to fetch the current username.  I've added a comment indicating that.

daschl:
Updated by a clean #216

PR #220, opened 2020-05-21
Adding example of how to run an adhoc query using SPC4.x

deniswsrosa:
Ad hoc Queries are not executed on the Cluster, I updated the migration guide to add a quick example of how it was in SPC3.x and how the new syntax looks like on SPC4.x. 

A proper session for the Cluster object will be added in the future

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).

### DIFF
--- a/src/main/asciidoc/migrating.adoc
+++ b/src/main/asciidoc/migrating.adoc
@@ -154,6 +154,58 @@ public class MyService {
 See <<couchbase.repository>> for more information.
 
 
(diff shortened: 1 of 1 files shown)


daschl:
Closing after discussion with @deniswsrosa because the raw queries are now possible in 4.1

PR #221, opened 2020-05-29
DATACOUCH-524 - ability to store @Nullable fields as field:null

mikereiche:
…the field is Optional.empty()

MOTIVATION
==========
This was a customer request.
Store Optional<T> fields as field:null if the field is Optional.empty().
If the field is null, do not emit the field at all (normal handling)

CHANGES
=======
Use the Jdk8Module of the ObjectMapper to handle Optional<T>.
In MappingCouchbaseConverter, recognize Optional<T> fields when reading and writing

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
@@ -39,6 +39,7 @@ import org.springframework.data.couchbase.core.mapping.CouchbaseList;
 import org.springframework.data.couchbase.core.mapping.CouchbaseMappingContext;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
+import org.springframework.data.couchbase.core.mapping.event.AfterConvertCallback;
 import org.springframework.data.couchbase.core.mapping.id.GeneratedValue;
 import org.springframework.data.couchbase.core.mapping.id.IdAttribute;
 import org.springframework.data.couchbase.core.mapping.id.IdPrefix;
@@ -458,8 +459,10 @@ public class MappingCouchbaseConverter extends AbstractCouchbaseConverter implem
 			return (String) propertyObj;
 		} else if (propertyObj instanceof Number) {
 			return new StringBuffer().append(propertyObj).toString();
-		} else {
+		} else if (propertyObj != null) {
 			return propertyObj.toString();
+		} else {
+			return null;
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


mp911de:
Please don't. Fields typed to `Optional` are discouraged. `Optional` should be used only at the method level for a proper domain model.

daschl:
@mp911de what's the best practice on distinguishing between a null value and one that is missing?

mp911de:
This depends a bit on the store. Here are a few examples:

* MongoDB can represent `null` but we decided to not write `null` fields and no one really asks for adding fields that contain a `null` value. Every save creates always a full document so previously existing values (and null values) get replaced by a fresh document.
* Cassandra: For Cassandra inserts, we omit `null` values as they create tombstones that need to be cleaned up. For updates, we set columns to `null` to reflect the object state.

mikereiche:
> Please don't. Fields typed to `Optional` are discouraged. `Optional` should be used only at the method level for a proper domain model.

Ok.  I totally agree that this would be a mis-use of Optional.   The customer did not ask for that, that was my idea.  It was the only idea I had at the time to distinguish among valued/null/missing at the domain-level.  @Nullable would be more appropriate - that would limit a @Nullable field to be either valued or 'null', with non-annotated fields to be either valued or missing.

mikereiche:
I've removed the special handling for Optional<>. Instead, @Nullable will cause field:null to be written to the database.   Our JacksonTranslationService has to be able to deal with such null values. 

The KUK code formatting definition I am using seems to differ from the code formatting that was used in the past (see MappingCouchbaseConverter).  Is there a different formatting definition I should be using)?

mp911de:
Thanks Mike. The `@Nullable` annotation is an interesting path to explore. We have a similar approach in other modules with embeddable objects. Please use the Spring Data formatting conventions. You can find a IDE guide along with the formatter settings at: https://github.com/spring-projects/spring-data-build/tree/master/etc/ide

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -501,33 +507,33 @@ public void doWithPersistentProperty(final CouchbasePersistentProperty prop) {
...
-						idAttributes.put(order, convertToString(propertyObj));
-					}
-
-					if (!conversions.isSimpleType(propertyObj.getClass())) {
-						writePropertyInternal(propertyObj, target, prop);
-					} else {
-						writeSimpleInternal(propertyObj, target, prop.getFieldName());
-					}
+
+				if (prop.isAnnotationPresent(IdPrefix.class)) {
```
mp911de, line 511: Please note: Annotation-based information should go inside of `PersistentProperty`. That will allow us in the future to react better to reflection-free metadata-caching.

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -501,33 +507,33 @@ public void doWithPersistentProperty(final CouchbasePersistentProperty prop) {
...
+					return;
+				}
+
+				if (prop.isAnnotationPresent(IdAttribute.class)) {
+					IdAttribute idAttribute = prop.findAnnotation(IdAttribute.class);
+					int order = idAttribute.order();
+					idAttributes.put(order, convertToString(propertyObj));
 				}
+
+				if (propertyObj != null && !conversions.isSimpleType(propertyObj.getClass())) {
```
mp911de, line 531: Nullability-annotations can be tricky. There are JSR-305 and Spring Framework variants of which Spring's annotations are meta-annotated with the JSR-305 ones. We have already nullability utilities in place (see [`o.s.d.util.NullableUtils`](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/util/NullableUtils.java).

The nullability lookup should also go into `CouchbasePersistentProperty`. We should consider extending `NullableUtils` for now to check whether a property is explicitly marked as nullable. Right now, the utilities only consider `MethodParameter` which is applicable for property accessors but not fields.

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
mikereiche: It's still inside doWithPersistentPropery().  I've just removed the if(propertyObj == null) { ... } guard around it, because it might need to be saved even if it is null.  That's what causes it to be out-dented.

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
mikereiche: I have an isNullable() in CouchbasePersistentProperty now.  It is referenced in the 'else'.   (I think I need to re-push).

				if (propertyObj != null && !conversions.isSimpleType(propertyObj.getClass())) {
					writePropertyInternal(propertyObj, target, prop);
				} else if (propertyObj != null || prop.isNullable()) {
					writeSimpleInternal(propertyObj, target, prop.getFieldName());
				}

mikereiche:
I've re-pushed this.  Please let me know if it is sufficient for now like this. It can be revisited if there are enhancements to NullableUtils.

PR #223, opened 2020-06-02
DATACOUCH-525 - support custom type field name and custome type field…

mikereiche:
… value

MOTIVATION
customers would like to specify a type field other than _class, and
would like to specify a type field value other than the class name
of the domain object

CHANGES
Follow the instructions at
https://stackoverflow.com/questions/38847605/how-to-use-spring-data-with-couchbase-without-class-attribute
in the test domain, supply a CustomMappingCouchbaseConvert class that leverates a
TypeBasedCouchbaseTypeMapper, which in turn uses a TypeAwareTypeInformationMapper.
The typeKey used is "t". The TypeAwareTypeInformationMapper uses the DocumentType annotation
as the value for the type map entry. The domain Person object has @DocumentType("p"), which
will result in type mapping of "t" : "p" for Person objects.

In addition, when a query is composed, the typeKey and typeValue are taken from the converter.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseConverter.java
@@ -21,6 +21,8 @@ import org.springframework.data.convert.EntityReader;
 import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
+import org.springframework.data.mapping.Alias;
+import org.springframework.data.util.TypeInformation;
 
 /**
  * Marker interface for the converter, identifying the types to and from that can be converted.
@@ -54,4 +56,9 @@ public interface CouchbaseConverter
 	 * @return the name of the field that will hold type information.
 	 */
 	String getTypeKey();
+
+	/**
+	 * @return the alias value for the type
+	 */
+	Alias getTypeAlias(TypeInformation<?> info);
 }
--- a/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseTypeMapper.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseTypeMapper.java
@@ -18,14 +18,19 @@ package org.springframework.data.couchbase.core.convert;
 
 import org.springframework.data.convert.TypeMapper;
 import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
+import org.springframework.data.mapping.Alias;
+import org.springframework.data.util.TypeInformation;
 
 /**
  * Marker interface for the TypeMapper.
  *
  * @author Michael Nitschinger
+ * @author Michael Reiche
  */
 public interface CouchbaseTypeMapper extends TypeMapper<CouchbaseDocument> {
 
 	String getTypeKey();
+
+	Alias getTypeAlias(TypeInformation<?> info);
 
 }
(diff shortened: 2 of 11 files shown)


src/main/java/org/springframework/data/couchbase/core/mapping/DocumentType.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Michael Nitschinger
+ * @author Andrey Rubtsov
+ * @author Michael Reiche
+ */
+@Persistent
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface DocumentType {
```
daschl, line 35: Use TypeAlias as talked about offline

src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseTypeMapper.java
```diff
@@ -28,4 +30,6 @@
 
 	String getTypeKey();
 
+	Alias getAlias(TypeInformation<?> info);
```
daschl, line 33: same as above

src/main/java/org/springframework/data/couchbase/core/query/Query.java
```diff
@@ -261,6 +264,11 @@ public String toN1qlString(ReactiveCouchbaseTemplate template, Class domainClass
 				domainClass);
 		MappingCouchbaseEntityInformation<?, Object> info = new MappingCouchbaseEntityInformation<>(persistentEntity);
 		String typeValue = info.getJavaType().getName();
+		TypeInformation<?> typeInfo= ClassTypeInformation.from(info.getJavaType());
```
daschl, line 267: formatting looks off

src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseTypeMapper.java
mikereiche: ok. fixed in all places.

src/main/java/org/springframework/data/couchbase/core/mapping/DocumentType.java
mikereiche: fixed

src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseConverter.java
```diff
@@ -54,4 +56,9 @@
 	 * @return the name of the field that will hold type information.
 	 */
 	String getTypeKey();
+
+	/**
+	 * @return the alias value for the type
+	 */
+	Alias getAlias(TypeInformation<?> info);
```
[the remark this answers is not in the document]
mikereiche: fixed 
(the reason I used getAlias instead of getTypeAlias is because the implementation relies on a method org.springframework.data.convert.DefaultTypeMapper.getAliasFor(type) (not getTypeAliasFor) )   Although in that method it is clear from the argument that it is an alias for the type.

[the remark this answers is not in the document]
daschl: so the method below is addressed, but not this one?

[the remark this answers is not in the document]
daschl: other than that, looks good to me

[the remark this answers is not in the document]
mikereiche: I refactored the one in CouchbaseTypeMapper and thought that was it. 
I've changed the definition here in CouchbaseConverter, the implementation in MappingCouchbaseConverter and use in Query.

PR #225, opened 2020-06-04
DATACOUCH-545 - return null if not found instead of throwing an excep…

mikereiche:
…tion

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
@@ -50,8 +50,8 @@ public class ReactiveFindByQueryOperationSupport implements ReactiveFindByQueryO
 		private final Query query;
 		private final QueryScanConsistency scanConsistency;
 
-		ReactiveFindByQuerySupport(final ReactiveCouchbaseTemplate template, final Class<T> domainType,
-				final Query query, final QueryScanConsistency scanConsistency) {
+		ReactiveFindByQuerySupport(final ReactiveCouchbaseTemplate template, final Class<T> domainType, final Query query,
+				final QueryScanConsistency scanConsistency) {
 			this.template = template;
 			this.domainType = domainType;
 			this.query = query;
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperationSupport.java
```diff
@@ -53,7 +54,12 @@
 
 		@Override
 		public T one(final String id) {
-			return reactiveSupport.one(id).block();
+			try {
+				return reactiveSupport.one(id).block();
+			} catch(DocumentNotFoundException dnfe){
+				dnfe.printStackTrace();
```
daschl, line 60: this is a leftover I assume?

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
```diff
@@ -70,7 +70,7 @@ public ReactiveFindByQueryOperationSupport(final ReactiveCouchbaseTemplate templ
 
 		@Override
 		public Mono<T> one() {
-			return all().single();
+			return all().singleOrEmpty();
```
daschl, line 73: If this is changed here, does it also need to be changed for analytics?

mikereiche: Yes, it probably needs to be changed there as well. I'll take a look.  I'll look for 'single()' everywhere.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperationSupport.java
mikereiche: I think it needs to be here in addition to ReactiveFindByQueryOperationSupport.  I had myself convinced there are different code-paths.  I'll check again.

daschl: Sorry, I meant the printStackTrace will dump into the logs and seems to be a leftover of debugging

PR #226, opened 2020-06-04
DATACOUCH-558 - Write Nested Objects as Map

mikereiche:
This is to fix an issue where nested objects were written as a
nested CouchbaseDocument - complete with id, expiration and
content containing the actual data. Aside from not being properly
stored, the documents could not be read back to construct a
domain object.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/domain/Address.java
+++ b/src/test/java/org/springframework/data/couchbase/domain/Address.java
@@ -1,0 +1,43 @@
+package org.springframework.data.couchbase.domain;
+
+import org.springframework.data.annotation.Id;
+import org.springframework.data.couchbase.config.AbstractCouchbaseConfiguration;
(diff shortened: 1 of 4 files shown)


src/test/java/org/springframework/data/couchbase/domain/Address.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.data.annotation.Id;
+import org.springframework.data.couchbase.config.AbstractCouchbaseConfiguration;
+import org.springframework.data.couchbase.core.mapping.Document;
+import org.springframework.data.couchbase.core.mapping.id.GeneratedValue;
+import org.springframework.data.couchbase.core.mapping.id.GenerationStrategy;
+
+import java.util.UUID;
+
+@Document
+public class Address extends AbstractEntity {
```
daschl, line 12: I see the entity changes, but can you add a automated test case for this so we catch any regressions?

PR #227, opened 2020-06-05
DATACOUCH-560 - serialize CouchbaseDocument and CouchbaseList

mikereiche:
MOTIVATION
==========
Without this change, List<> and Map<> are just serialized as beans
A non-empty list ends up being { empty : false }
Also included are MappingCouchbaseConverter and CouchbaseDocument
from DATACOUCH-558 as they are needed for the manual testing of
Person.Address.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -22,6 +22,10 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
+import com.couchbase.client.java.codec.JacksonJsonSerializer;
+import com.couchbase.client.java.json.JsonValueModule;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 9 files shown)


src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -140,7 +145,22 @@ public ClusterEnvironment couchbaseClusterEnvironment() {
 	 * @param builder the builder that can be customized.
 	 */
 	protected void configureEnvironment(final ClusterEnvironment.Builder builder) {
+		Jackson2ObjectMapperBuilder jacksonBuilder = new Jackson2ObjectMapperBuilder();
```
daschl, line 148: Actually I don't think this is how it should work? We do not expose jackson directly to the user, and all the encoding and decoding happens on the spring side and not with the SDK itself.. shouldn't this go into the mapper/converter logic instead?

daschl: pinging @mp911de

PR #228, opened 2020-06-08
DATACOUCH-560 - Serialization of CouchbaseDocument and CouchbaseList.

runnerway:
Add support for saving document with lists and composed objects.
Convert CouchbaseDocument/CouchbaseList to Map/List before saving.

There is another PR for the same ticket (https://github.com/spring-projects/spring-data-couchbase/pull/227) but i would like to propose another approach to solve the problem.


mikereiche:
I will add tests for this in a separate merge.

PR #232, opened 2020-06-18
DATACOUCH-546 - set generated id on source entity to be returned to c…

mikereiche:
DATACOUCH-546 - set generated id on source entity to be returned to caller

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
@@ -18,21 +18,18 @@ package org.springframework.data.couchbase.core.convert;
 
 import static org.springframework.data.couchbase.core.mapping.id.GenerationStrategy.*;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -959,10 +947,190 @@ public ConverterAwareSpELExpressionParameterValueProvider(final SpELExpressionEv
...
+	 * 3) X id -> source.setId(Y.newInstance(X))
+	 * 4) X id -> source.setId(Y.from(X));
+	 * 5) X id -> source.setId(Y.newInstance(X.toString()) this will only work if T.toString() is appropriate
+	 * 6) X id -> source.setId(Y.fromString(X))  this will only work if T.toString() is appropriate
+	 * If none of that works, try to set the id or @Id field directly
+	 *
+	 * @param source object being converted
+	 * @param id     id that was used in the target
+	 */
+	private static void setId(Object source, Object id) {
```
mp911de, line 969: That's not how it's supposed to work. Spring Data provides a `PersistentEntity`/`PersistentProperty` meta-model to identify ID properties. Property values are read and written through `PersistentPropertyAccessor` that ensure proper handling of immutable objects (that is creating a new instance for Kotlin data classes and returning the modified instance).

PR #241, opened 2020-07-01
DATACOUCH-533 - Promote id field if no annotation id.

mikereiche:
Promote id field if no annotation id.
Note that there is no longer any Couchbase @Id annotation - only
the Spring @id annotation.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
+++ b/src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
@@ -23,6 +23,8 @@ import java.util.concurrent.TimeUnit;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.core.env.Environment;
 import org.springframework.data.annotation.Id;
+import org.springframework.data.mapping.MappingException;
+import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.model.BasicPersistentEntity;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.util.Assert;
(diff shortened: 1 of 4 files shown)


mp911de:
Does Couchbase default to `id` as field name for the primary key in each document?

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
```diff
@@ -97,6 +97,6 @@ public String getFieldName() {
 	// DATACOUCH-145: allows SDK's @Id annotation to be used
 	@Override
 	public boolean isIdProperty() {
-		return isAnnotationPresent(Id.class) || super.isIdProperty();
+		return isAnnotationPresent(Id.class) || super.isIdProperty() || this.getFieldName().equals("id");
```
daschl, line 100: maybe check with lower case to be more resilient?

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
```diff
@@ -76,24 +78,22 @@ protected CouchbasePersistentProperty returnPropertyIfBetterIdPropertyCandidateO
...
-			// spring IDs will have priority over SDK IDs
+		if (candidateSpringId && !currentSpringId) {
+			// spring IDs will have priority over fields named id
 			return property;
-		} else if (currentSpringId && candidateCbId) {
-			// ignore SDK's IDs if current is a Spring ID
+		} else if (currentSpringId && !candidateSpringId) {
+			// spring IDs will have priority over fields named id
 			return null;
+		} else { // if (currentSpringId && candidateSpringId || !currentSpringId && !candidateSpringId)
```
daschl, line 90: this comment is confusing on the else

src/test/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentPropertyTests.java
```diff
@@ -82,6 +83,93 @@ void testSdkIdAnnotationEvaluatedAfterSpringIdAnnotationIsIgnored() {
 		assertThat(test.getIdProperty()).isEqualTo(springIdProperty);
 	}
 
+	@Test
+	void testIdFieldOnly() {
+		{ // only has @springId
```
daschl, line 88: can you split these up into individual unit tests?

daschl:
> Does Couchbase default to `id` as field name for the primary key in each document?

@mp911de not sure I understand the question. Right now we require the Id annotation to be present on a field, but it makes sense to imply it if there is no Id annotation but an id field.

mp911de:
Let me add a bit of context here. We have this kind of functionality only in MongoDB where Id fields map to a field named `_id` in the MongoDB document (`{_id : <primary key goes here>}`). This is to simplify the process as everyone would need to add a `_id` Java property. All other modules do not use this convention to not get into the way of domain models that name a property `id` but it's not the primary key.

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
mikereiche: With two binary variables, there are four cases.   the 1 0 and 0 1 cases are above.  This else is the 0 0 and 1 1 cases.
1 1 is two fields with @Id annotation and is not allowed.  0 0 is two fields named id - which is possible through the @Field annotation.

src/test/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentPropertyTests.java
mikereiche: Done

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
mikereiche: done

mikereiche:
> Let me add a bit of context here. We have this kind of functionality only in MongoDB where Id fields map to a field named `_id` in the MongoDB document (`{_id : <primary key goes here>}`). This is to simplify the process as everyone would need to add a `_id` Java property. 

Couchbase was using the @Id annotation to specify which field is the ID (getId()) for the entity.  The presence of the @Id annotation indicates that this field will be used in the document-id (the document-id may also contain prefixes, suffixes and other fields (attributes) from the entity).  The field that has the @Id annotation is not included in the document as it is already included as the document-id (unlike attribute fields, which are in the document-id, and also in the document). 
This change is in response to users comments "I have an 'id' field, why does CB require that I put an @Id annotation on it?" - so if there is no other @Id field, the field named 'id' is used as the @Id field.  There are checks against having two @Id fields or two fields named 'id' (which is possible via the @Field("name") annotation). If there is a field with the @Id annotation, that field has precedence over a field named 'id'

> All other modules do not use this convention to not get into the way of domain models that name a property `id` but it's not the primary key.

Couchbase will still respect this, provided that the other field has an @Id annotation - which was required anyway, as Couchbase requires that the target.getId() is not null.

PR #245, opened 2020-07-14
DATACOUCH-580 - Do not automatically insert n1ql.filter predicate in …

mikereiche:
…string query.

Since it is difficult to determine where the n1ql.filter (which typically
filters on _class), and also because it is not mandatory, including
the n1ql.filter in the @Query is left up to the author of the query.
Example:
   @Query("#{#n1ql.selectEntity} where #{#n1ql.filter} and lastname = $1")

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
@@ -21,11 +21,16 @@ import org.springframework.data.couchbase.core.ReactiveCouchbaseTemplate;
 import org.springframework.data.couchbase.repository.query.StringBasedN1qlQueryParser;
 
 /**
- * Query created from the string in @Query annotation in the repository interface. @Query("#{#n1ql.selectEntity} where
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
```diff
@@ -21,16 +21,12 @@
 import org.springframework.data.couchbase.repository.query.StringBasedN1qlQueryParser;
 
 /**
- * @author Michael Reiche
- *
- * Query created from the string in @Query annotation in the repository interface.
- *
- *      @Query("#{#n1ql.selectEntity} where firstname = $1 and lastname = $2")
- * 	    List<User>  getByFirstnameAndLastname(String firstname, String lastname);
+ * Query created from the string in @Query annotation in the repository interface. @Query("#{#n1ql.selectEntity} where
```
dnault, line 24: What's the Spring convention for including examples in Javadoc? Should the code snipped be enclosed in "pre" tags?

mikereiche: Fixed. enclosed in "pre"

PR #248, opened 2020-07-21
DATACOUCH-588 - Implement Pageable Queries for non-Reactive repositor…

mikereiche:
…ies.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/query/Query.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/Query.java
@@ -104,11 +104,20 @@ public class Query {
 	 * Limit the number of returned documents to {@code limit}.
 	 *
 	 * @param limit
-	 * @return
+	 * @return this
 	 */
 	public Query limit(int limit) {
 		this.limit = limit;
 		return this;
+	}
+
+	/**
+	 * limit
+	 *
+	 * @return limit
+	 */
+	public int getLimit() {
+		return limit;
 	}
 
 	/**
--- a/src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryExecution.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryExecution.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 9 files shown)


mikereiche:
This only has the implementation for non-Reactive repositories.  The implementation for Reactive is more complicated and will follow in a separate change.

src/main/java/org/springframework/data/couchbase/core/query/Query.java
```diff
@@ -104,13 +104,22 @@ public Query skip(long skip) {
...
 	 */
 	public Query limit(int limit) {
 		this.limit = limit;
 		return this;
 	}
 
+	/**
+	 * limit
+	 *
+	 * @return limit
```
mp911de, line 117: Please add `@since` tags to indicate when a particular class/method was introduced. This is helpful to users to know since when they can use a certain functionality.

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryExecution.java
```diff
@@ -0,0 +1,172 @@
...
+import org.springframework.data.domain.Slice;
+import org.springframework.data.domain.SliceImpl;
+import org.springframework.data.repository.support.PageableExecutionUtils;
+import org.springframework.util.Assert;
+
+/**
+ * Set of classes to contain query execution strategies. Depending (mostly) on the return type of a
+ * {@link org.springframework.data.repository.query.QueryMethod} a {@link Query} can be executed in various flavors.
+ *
+ * @author Oliver Gierke
```
mp911de, line 33: No need to keep the original authors on copied patterns unless they are exactly the same.

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -135,6 +138,11 @@ void count() {
 				airportRepository.save(airport);
 			}
 
+			Pageable pageable =  PageRequest.of(0, 2);
+			Page<Airport> aPage = 		airportRepository.findAllByIataNot("JFK", pageable);
```
mp911de, line 142: Nit: Formatting

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -108,6 +112,11 @@ void count() {
 				airportRepository.save(airport).block();
 			}
 
+			// Pageable pageable = PageRequest.of(0, 2);
```
mp911de, line 115: We shouldn't add commented out code blocks.

PR #251, opened 2020-07-28
DATACOUCH-595 - Add cas to replace options if present on entity.

mikereiche:

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
@@ -96,6 +96,21 @@ class CouchbaseTemplateSupport implements ApplicationContextAware {
 		}
 	}
 
+	public long getCas(final Object entity) {
+		final ConvertingPropertyAccessor<Object> accessor = getPropertyAccessor(entity);
+		final CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(entity.getClass());
+		final CouchbasePersistentProperty versionProperty = persistentEntity.getVersionProperty();
+
+		long cas = 0;
+		if (versionProperty != null) {
+			Object casObject  =  (Number)accessor.getProperty(versionProperty);
+			if (casObject instanceof Number){
+				cas = ((Number)casObject).longValue();
+			}
+		}
+		return cas;
+	}
+
 	public String getJavaNameForEntity(final Class<?> clazz) {
 		final CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(clazz);
 		MappingCouchbaseEntityInformation<?, Object> info = new MappingCouchbaseEntityInformation<>(persistentEntity);
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
@@ -68,8 +68,8 @@ public class ReactiveReplaceByIdOperationSupport implements ReactiveReplaceByIdO
 			return Mono.just(object).flatMap(o -> {
 				CouchbaseDocument converted = template.support().encodeEntity(o);
 				return template.getCollection(collection).reactive()
-						.replace(converted.getId(), converted.export(), buildReplaceOptions()).map(result -> {
-							template.support().applyUpdatedCas(object, result.cas());
+						.replace(converted.getId(), converted.export(), buildReplaceOptions(o)).map(result -> {
+							template.support().applyUpdatedCas(o, result.cas());
 							return o;
 						});
 			}).onErrorMap(throwable -> {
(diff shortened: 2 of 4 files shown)


src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -96,6 +96,18 @@ public void applyUpdatedCas(final Object entity, final long cas) {
 		}
 	}
 
+	public long getCas(final Object entity) {
```
mp911de, line 99: Nit: We typically don't use `final` keywords within methods.

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -96,6 +96,18 @@ public void applyUpdatedCas(final Object entity, final long cas) {
...
+		final ConvertingPropertyAccessor<Object> accessor = getPropertyAccessor(entity);
+		final CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(entity.getClass());
+		final CouchbasePersistentProperty versionProperty = persistentEntity.getVersionProperty();
+
+		long cas = 0;
+		if (versionProperty != null) {
+			cas = (long) accessor.getProperty(versionProperty);
```
mp911de, line 106: Probably safer to have an instanceof check for `Number` as domain models may contain `BigInteger` and other numeric types.

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateKeyValueIntegrationTests.java
```diff
@@ -74,6 +74,12 @@ void upsertAndFindById() {
 		User modified = couchbaseTemplate.upsertById(User.class).one(user);
 		assertEquals(user, modified);
 
+		modified = couchbaseTemplate.replaceById(User.class).one(user);
+		assertEquals(user, modified);
```
mp911de, line 78: On a related note, Spring Data has switched to AssertJ which makes assertions a bit more convenient.

src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
```diff
@@ -68,7 +68,7 @@ public ReactiveReplaceByIdOperationSupport(final ReactiveCouchbaseTemplate templ
 			return Mono.just(object).flatMap(o -> {
 				CouchbaseDocument converted = template.support().encodeEntity(o);
 				return template.getCollection(collection).reactive()
-						.replace(converted.getId(), converted.export(), buildReplaceOptions()).map(result -> {
+						.replace(converted.getId(), converted.export(), buildReplaceOptions(object)).map(result -> {
```
daschl, line 71: nit: this should be o, not object, right?

daschl: actually I wonder if the code from below is put up here and only the long cas is put into the buildReplaceOptions

src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
```diff
@@ -86,13 +86,17 @@ public ReactiveReplaceByIdOperationSupport(final ReactiveCouchbaseTemplate templ
...
-		private ReplaceOptions buildReplaceOptions() {
+		private ReplaceOptions buildReplaceOptions(T object) {
 			final ReplaceOptions options = ReplaceOptions.replaceOptions();
 			if (persistTo != PersistTo.NONE || replicateTo != ReplicateTo.NONE) {
 				options.durability(persistTo, replicateTo);
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
+			long cas = template.support().getCas(object);
+			if (cas != 0) {
```
daschl, line 97: 0 cas == not present so if 0 is passed in its fine

src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
mikereiche: Yes. should be o.  Fixed.

src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
mikereiche: fixed

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
mikereiche: fixed

src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
mikereiche: The durability also needs to go in the options.

daschl (approved):

daschl:
@mikereiche also I think this should be backported to 4.0 ?

PR #260, opened 2020-09-14
DATACOUCH-604 - Fix NPE when looking for Id field in interface.

mikereiche:

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
+++ b/src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
@@ -16,7 +16,6 @@
 
 package org.springframework.data.couchbase.core.mapping;
 
-import org.springframework.data.annotation.Id;
 import org.springframework.data.mapping.Association;
 import org.springframework.data.mapping.MappingException;
 import org.springframework.data.mapping.model.AnnotationBasedPersistentProperty;
@@ -27,8 +26,6 @@ import org.springframework.data.mapping.model.SimpleTypeHolder;
 import org.springframework.util.StringUtils;
 
 import com.couchbase.client.core.deps.com.fasterxml.jackson.annotation.JsonProperty;
-
-import java.util.Locale;
 
 /**
  * Implements annotated property representations of a given {@link Field} instance.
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
```diff
@@ -28,6 +28,7 @@
 
 import com.couchbase.client.core.deps.com.fasterxml.jackson.annotation.JsonProperty;
 
+import java.security.acl.LastOwnerException;
```
daschl, line 31: unneeded import?

daschl (changes requested):
Can we add  a test case as a regression test for this npe?

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
```diff
@@ -99,7 +100,13 @@ public String getFieldName() {
 	// DATACOUCH-145: allows SDK's @Id annotation to be used
 	@Override
 	public boolean isIdProperty() {
-		return isAnnotationPresent(Id.class) || super.isIdProperty()
-				|| this.getFieldName().toLowerCase(Locale.ROOT).equals("id");
+		if (isAnnotationPresent(Id.class) || super.isIdProperty()){
```
mp911de, line 103: This check can be simplified to:

```java
		if (super.isIdProperty()) {
			return true;
		}
```

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
```diff
@@ -99,7 +100,13 @@ public String getFieldName() {
...
+			return true;
+		}
+		// is field named "id"
+		if(getField() != null && this.getFieldName().toLowerCase(Locale.ROOT).equals("id")){
```
mp911de, line 107: In MongoDB, we simplify that check to:

```
SUPPORTED_ID_PROPERTY_NAMES.contains(getName()) && !hasExplicitFieldName()
```

Here, you could check for e.g. `"id".equals(getName())`.

mp911de (commented):
I left two comments how the check could be simplified. Especially `isAnnotationPresent(…)` shouldn't be necessary.

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
mikereiche: Done. Thanks!

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
mikereiche: Done. Thanks!

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentProperty.java
mikereiche: Removed.  I have no idea.

daschl (approved):

PR #263, opened 2020-09-17
DATACOUCH-603 Do not cast query parameters in N1qlQueryCreator.

mikereiche:
For IN and NOT_IN - they can take varargs, and array or a JsonArray
- don't cast the parameter, let parameter accessor handle that.
- JsonArray.toString() provides the brackets, don't include them in the format
- Object[] array needs explicit conversion, as toString() is not sufficient

Also cleaned up some weirdness in QueryCriteria.


src/test/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreatorTests.java
```diff
@@ -62,6 +63,35 @@ void createsQueryCorrectly() throws Exception {
 		assertEquals(query.export(), " WHERE " + where("firstname").is("Oliver").export());
 	}
 
+	@Test
```
daschl, line 66: Would it make sense to add another test for List?

daschl (changes requested):
looks good overall, just one note on a test

daschl (approved):

PR #264, opened 2020-09-21
DATACOUCH-615 Use query in removeByQuery

haal:
Use provided query instead of hardcoded statement.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
@@ -58,14 +58,7 @@ public class ReactiveRemoveByQueryOperationSupport implements ReactiveRemoveByQu
 		@Override
 		public Flux<RemoveResult> all() {
 			return Flux.defer(() -> {
-				String bucket = "`" + template.getBucketName() + "`";
-
-				String typeKey = template.getConverter().getTypeKey();
-				String typeValue = template.support().getJavaNameForEntity(domainType);
-				String where = " WHERE `" + typeKey + "` = \"" + typeValue + "\"";
-
-				String returning = " RETURNING meta().*";
-				String statement = "DELETE FROM " + bucket + " " + where + returning;
+				String statement = assembleDeleteQuery();
 
 				return template.getCouchbaseClientFactory().getCluster().reactive().query(statement, buildQueryOptions())
 						.onErrorMap(throwable -> {
@@ -75,7 +68,7 @@ public class ReactiveRemoveByQueryOperationSupport implements ReactiveRemoveByQu
 								return throwable;
 							}
 						}).flatMapMany(ReactiveQueryResult::rowsAsObject)
-						.map(row -> new RemoveResult(row.getString("id"), row.getLong("cas"), Optional.empty()));
+						.map(row -> new RemoveResult(row.getString("__id"), row.getLong("__cas"), Optional.empty()));
 			});
 		}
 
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@haal Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=264)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=264) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@haal Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=264)!

src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
```diff
@@ -75,7 +68,7 @@ public ReactiveRemoveByQueryOperationSupport(final ReactiveCouchbaseTemplate tem
 								return throwable;
 							}
 						}).flatMapMany(ReactiveQueryResult::rowsAsObject)
-						.map(row -> new RemoveResult(row.getString("id"), row.getLong("cas"), Optional.empty()));
+						.map(row -> new RemoveResult(row.getString("__id"), row.getLong("__cas"), Optional.empty()));
```
mikereiche, line 71: The names (__id, __cas) should come from TemplateUtils, although they seem to be hard-coded in several other places.

src/main/java/org/springframework/data/couchbase/core/query/Query.java
```diff
@@ -254,6 +254,16 @@ public String toN1qlString(ReactiveCouchbaseTemplate template, Class domainClass
 		return statement.toString();
 	}
 
+	public String toN1qlRemoveString(ReactiveCouchbaseTemplate template, Class domainClass) {
+		StringBasedN1qlQueryParser.N1qlSpelValues n1ql = getN1qlSpelValues(template, domainClass, false);
+		final StringBuilder statement = new StringBuilder();
+		appendString(statement, n1ql.delete); // delete ...
+		appendWhereString(statement, n1ql.filter); // typeKey = typeValue
+		appendWhere(statement, null); // criteria on this Query
```
mikereiche, line 262: If the second arg to appendWhere() is  new int[] { 0 }   (as in toN1qlSelectString), the variables in the criteria will be parameterized for efficiency.

mikereiche (approved):

src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
haal: Fixed

src/main/java/org/springframework/data/couchbase/core/query/Query.java
haal: Initially I had the same idea but in that case positional parameters in query will not be replaced with actual values. Also, it seems in current code `findByQuery(clazz).matching(query)` method is not behaving as expected (added new test case in `CouchbaseTemplateQueryIntegrationTests` in this PR which is failing).

daschl: I don't think we should submit a pr with a failing test - maybe submit it in a different PR so we can work on a fix alongside it?

haal: Yes, I can create failing test in separate PR. The question is how to handle comment from @mikereiche where he proposes to use `new int[] {0}` which does not work at the moment?

haal: New bug created https://jira.spring.io/browse/DATACOUCH-619. 
Resolving initial comment without applying `new int[] {0}` as it is not working as expected.

daschl (approved):

PR #265, opened 2020-09-22
DATACOUCH-616 - Fix parsing of String Query containing conditionals i…

mikereiche:
…n quotes

For a query as below that has conditional portions, the parsing for
parameters was being done before the conditional portions were being
resolved, which left the parameters between quotes where they wre
not being recognized as query parameters.

@Query("#{#n1ql.selectEntity} WHERE #{#n1ql.filter} " +
	" #{#projectIds != null ? 'AND iata IN $1' : ''} ")
Long count(@Param("projectIds") List<String> projectIds)


src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -192,11 +191,11 @@ private PlaceholderType checkPlaceholders(String statement) {
 		}
 
 		if (posCount > 0) {
-			return PlaceholderType.POSITIONAL;
+			return placeHolderType = PlaceholderType.POSITIONAL;
```
daschl, line 194: assignment and return in one ?

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -103,7 +103,7 @@
 	private static final Logger LOGGER = LoggerFactory.getLogger(StringBasedN1qlQueryParser.class);
 	private final String statement;
 	private final QueryMethod queryMethod;
-	private final PlaceholderType placeHolderType;
+	private PlaceholderType placeHolderType;
```
daschl, line 106: you sure this is non final now? if so it maybe shold be volatile

src/main/java/org/springframework/data/couchbase/repository/query/StringN1qlQueryCreator.java
```diff
@@ -153,9 +153,10 @@ private QueryCriteria from(final Part part, final CouchbasePersistentProperty pr
 	// copied from StringN1qlBasedQuery
 	private N1QLExpression getExpression(ParameterAccessor accessor, Object[] runtimeParameters,
 			ReturnedType returnedType) {
-		EvaluationContext evaluationContext = evaluationContextProvider.getEvaluationContext(
-				getQueryMethod().getParameters(), runtimeParameters);
-		N1QLExpression parsedStatement = x(this.queryParser.doParse(parser, evaluationContext, false));
+		boolean isCountQuery = queryMethod.getName().toLowerCase().startsWith("count"); // should be queryMethod.isCountQuery()
```
daschl, line 156: is this still a todo?

daschl (changes requested):
put in a couple comments which I dont think look right

JesusTheHun:
Is it possible to backport this fix to future `3.2.11` is any ?

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
mikereiche: Yes. It used to be called int the constructor as 
     this.placeHolderType = checkPlaceholders(...)
Now it is public and gets called by StringN1qlQueryCreator - but the placeHolderType still needs to be set. I suppose it could be changed to not return a value - it was private, nothing else could be calling it for that.

I'll change it.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
mikereiche: It gets set when StringBasedN1qlQueryParser.checkPlaceHolders() is called by StringN1qlQueryCreator.getExpression(). To make it final, the code in getExpression() would need to be executed in the constructor. It can be done.  Not sure that it is better.  See next push.

src/main/java/org/springframework/data/couchbase/repository/query/StringN1qlQueryCreator.java
mikereiche: queryMethod.isCount() is in the datacouch-588 branch along with some others.

mikereiche:
> Is it possible to backport this fix to future `3.2.11` if any ?

Yes. The original issue was for 3.2.x

JesusTheHun:
I'm the author of the issue. I was not sure you were going to make a new release for the 3.x branch though ;)
Thx

src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
```diff
@@ -98,7 +97,7 @@ private InsertOptions buildInsertOptions() {
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
-			if (expiry != null) {
+			if (expiry != null && !expiry.equals(Duration.ofSeconds(0))) {
```
daschl, line 100: Nit: you can use Duration.ZERO;

src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
```diff
@@ -98,7 +98,7 @@ private UpsertOptions buildUpsertOptions() {
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
-			if (expiry != null) {
+			if (expiry != null && ! expiry.equals( Duration.ofSeconds(0))) {
```
daschl, line 101: Nit: you can use Duration.ZERO;

daschl (changes requested):
It would be good if we could have an integration test which covers this scenario, otherwise just minor nits

PR #267, opened 2020-10-02
DATACOUCH-605 - Support ScanConsistency in n1ql queries

mikereiche:

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.data.couchbase.core;
 
-import com.couchbase.client.java.kv.UpsertOptions;
 import org.springframework.data.couchbase.core.mapping.Document;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -98,7 +97,7 @@ public class ReactiveInsertByIdOperationSupport implements ReactiveInsertByIdOpe
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
-			if (expiry != null) {
+			if (expiry != null && ! expiry.isZero()) {
 				options.expiry(expiry);
 			} else if (domainType.isAnnotationPresent(Document.class)) {
 				Document documentAnn = domainType.getAnnotation(Document.class);
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
@@ -15,7 +15,6 @@
  */
 package org.springframework.data.couchbase.core;
 
-import com.couchbase.client.java.kv.Expiry;
 import org.springframework.data.couchbase.core.mapping.Document;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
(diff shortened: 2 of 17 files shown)


src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
```diff
@@ -98,7 +97,7 @@ private InsertOptions buildInsertOptions() {
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
-			if (expiry != null) {
+			if (expiry != null && ! expiry.isZero()) {
```
daschl, line 100: is this unrelated?

src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
```diff
@@ -98,7 +97,7 @@ private UpsertOptions buildUpsertOptions() {
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
-			if (expiry != null) {
+			if (expiry != null && !expiry.isZero()) {
```
daschl, line 100: is this unrelated?

src/main/java/org/springframework/data/couchbase/core/query/Query.java
```diff
@@ -127,6 +128,26 @@ public Query with(final Pageable pageable) {
 		return with(pageable.getSort());
 	}
 
+	/**
+	 * queryScanConsistency
+	 *
+	 * @return queryScanConsistency
+	 */
+	public QueryScanConsistency getConsistency() {
```
daschl, line 136: do we need a public getter on this?

src/main/java/org/springframework/data/couchbase/core/query/Query.java
```diff
@@ -127,6 +128,26 @@ public Query with(final Pageable pageable) {
...
+		return queryScanConsistency;
+	}
+
+	/**
+	 * Sets the given scan consistency on the {@link Query} instance.
+	 *
+	 * @param queryScanConsistency
+	 * @return this
+	 */
+	public Query with(final QueryScanConsistency queryScanConsistency) {
```
daschl, line 146: shouldn't this be more like scanConsistency(..) ? we also name limit etc. explicitly.

daschl (changes requested):
overall looks good, just a couple minor notes

src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
mikereiche: yes - unrelated. It's actually in the pr for datacouch-616.

src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
mikereiche: yes - unrelated. It's actually in the pr for datacouch-616.

src/main/java/org/springframework/data/couchbase/core/query/Query.java
mikereiche: It's used by ExecutableFindByQuerySupport.matching(Query) - if Query has scanConsistency set, it is taken from there instead of requiring a separate call to ExecutableFindByQuerySupport.consistentWith( scanConsistency ).

src/main/java/org/springframework/data/couchbase/core/query/Query.java
mikereiche: Ok - I changed it to scanConsistency().  But I was looking at with(Pageable) and with(Sort).
But that pattern breaks down for things like skip and limit which are just int.

PR #268, opened 2020-10-06
DATACOUCH-623 - Add replace() method to CouchbaseRepository for CAS u…

mikereiche:
…sage.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
@@ -29,6 +29,7 @@ import org.springframework.data.couchbase.core.ReactiveCouchbaseOperations;
 import org.springframework.data.couchbase.core.query.Query;
 import org.springframework.data.couchbase.repository.ReactiveCouchbaseRepository;
 import org.springframework.data.couchbase.repository.query.CouchbaseEntityInformation;
+import static org.springframework.data.couchbase.repository.support.Util.hasNonZeroVersionProperty;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.util.Streamable;
 import org.springframework.util.Assert;
@@ -75,9 +76,15 @@ public class SimpleReactiveCouchbaseRepository<T, ID> implements ReactiveCouchba
 	}
 
 	@SuppressWarnings("unchecked")
+	@Override
 	public <S extends T> Mono<S> save(final S entity) {
 		Assert.notNull(entity, "Entity must not be null!");
-		return (Mono<S>) operations.upsertById(entityInformation.getJavaType()).one(entity);
+		// if entity has non-null version property, then replace()
+		if (hasNonZeroVersionProperty(entity, operations.getConverter())) {
+			return (Mono<S>) operations.replaceById(entityInformation.getJavaType()).one(entity);
+		} else {
+			return (Mono<S>) operations.upsertById(entityInformation.getJavaType()).one(entity);
+		}
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


src/main/java/org/springframework/data/couchbase/repository/CouchbaseRepository.java
```diff
@@ -43,4 +43,8 @@
 	@Override
 	List<T> findAllById(Iterable<ID> iterable);
 
+	<S extends T> S replace(S var1);
```
mp911de, line 46: Adding technology-specific methods violates somewhat the concept of handling technology-specific aspects inside the repository/template API. I left a comment on DATACOUCH-623. Especially entities annotated with `@Version` shouldn't require the user to call a different method for proper handling, instead, the requirement to issue a different store-specific replace call should be determined by the Spring Data code and not the caller.

daschl:
@mikereiche after looking into this I agree with @mp911de's comment in https://jira.spring.io/browse/DATACOUCH-623?focusedCommentId=191380&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-191380 .. Let's not add to the repository, but rather in our internal code check if the version is present and if present and set use a replace operation and map to the corresponding exception.

src/main/java/org/springframework/data/couchbase/repository/CouchbaseRepository.java
mikereiche: Thanks Mark.  I put the upsert() vs. replace() logic in the save().

src/test/java/org/springframework/data/couchbase/domain/UserRepository.java
```diff
@@ -30,10 +32,14 @@
  * @author Michael Reiche
  */
 @Repository
-public interface UserRepository extends PagingAndSortingRepository<User, String> {
+public interface UserRepository extends CouchbaseRepository<User, String> {
 
 	List<User> findByFirstname(String firstname);
 
+	List<User> findByFirstnameIn(String... firstnames);
```
daschl, line 39: isn't this related to the other open pr?

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
```diff
@@ -77,7 +79,27 @@ public SimpleReactiveCouchbaseRepository(final CouchbaseEntityInformation<T, Str
...
+		// if entity has non-null version property, then replace()
+		if (hasVersionProperty(entity)) {
+			return (Mono<S>) operations.replaceById(entityInformation.getJavaType()).one(entity);
+		} else {
+			return (Mono<S>) operations.upsertById(entityInformation.getJavaType()).one(entity);
+
+		}
+	}
+
+	private boolean hasVersionProperty(Object entity) {
```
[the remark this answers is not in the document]
mikereiche: It can be done. I'll put it in a new Utils class.
I had thought of that, but  SimpleReactiveCouchbaseRepository has a ReactiveCouchbaseOperations, while SimpleCouchbaseRepository has a CouchbaseOperations.   But it doesn't actually need those - it just needs the mapping conext of the converter held by each *Operations - which is the same type.

src/test/java/org/springframework/data/couchbase/domain/UserRepository.java
mikereiche: Yes. Removed

mikereiche: I think this line will be back in after rebasing to master that has that change.

PR #272, opened 2020-10-13
DATACOUCH-626 - Handle reading into entity with only all-args constru…

mikereiche:
…ctor; version primative type

The version arg - which is not in the document - will be passed
into the constructor as a null which will fail if it is a primitive
type. This change will pre-populated the converted object with
the version/cas, so that it is not null.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/domain/SubscriptionToken.java
+++ b/src/test/java/org/springframework/data/couchbase/domain/SubscriptionToken.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2020 the original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.couchbase.domain;
+
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.ToString;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.annotation.Version;
+import org.springframework.data.couchbase.core.mapping.Document;
+import org.springframework.data.couchbase.core.mapping.Field;
+import org.springframework.data.couchbase.core.mapping.id.GeneratedValue;
+import org.springframework.data.couchbase.core.mapping.id.GenerationStrategy;
+
+/**
+ * SubscriptionTokenEntity for tests
+ *
+ * @author Michael Reiche
+ */
+@Getter
+@ToString
+@EqualsAndHashCode
+@Document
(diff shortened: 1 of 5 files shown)


src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryKeyValueIntegrationTests.java
```diff
@@ -95,31 +101,6 @@ void saveAndFindByIdWithList() {
 		assertFalse(userRepository.existsById(library.getId()));
 	}
 
-	@Test
```
daschl, line 98: why did this test get removed?

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryKeyValueIntegrationTests.java
```diff
@@ -54,7 +51,16 @@
...
 	@Autowired LibraryRepository libraryRepository;
-	@Autowired UserSubmissionRepository userSubmissionRepository;
+	@Autowired SubscriptionTokenRepository subscriptionTokenRepository;
+
+	@Test
+	void subscriptionToken(){
+		SubscriptionToken st = new SubscriptionToken("id", 0, "type", "Dave Smith",
+				"app123", "dev123", 0);
+		st = subscriptionTokenRepository.save(st);
+		st = subscriptionTokenRepository.findById(st.getId()).get();
```
daschl, line 61: can you assert that the cas stored in the token is the same as when read "raw" from the SDK to make sure it is populated properly?

src/test/java/org/springframework/data/couchbase/domain/SubscriptionTokenRepository.java
```diff
@@ -0,0 +1,16 @@
+package org.springframework.data.couchbase.domain;
```
daschl, line 1: license header missing?

src/test/java/org/springframework/data/couchbase/domain/SubscriptionToken.java
```diff
@@ -0,0 +1,54 @@
+package org.springframework.data.couchbase.domain;
```
daschl, line 1: license header missing?

daschl (changes requested):
Looks good to me, with couple nits

src/test/java/org/springframework/data/couchbase/domain/SubscriptionToken.java
mikereiche: Ok - will add.

src/test/java/org/springframework/data/couchbase/domain/SubscriptionTokenRepository.java
mikereiche: will add

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryKeyValueIntegrationTests.java
mikereiche: Ok, will add that check

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryKeyValueIntegrationTests.java
mikereiche: Oops. Bad editing. Will restore.

PR #274, opened 2020-10-15
DATACOUCH-625 - Save doesn't return updated entity if immutable.

mikereiche:
Modified applyUpdatedCas() to return accessor.getBean() which will
occur if the version property is Immutable.
Created applyUpdateId() that is analogous to applyUpdatedCas()
Factored out the equals() method that was used by a number of
test entity objects.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.couchbase.core;
 
+import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
+import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
 import org.springframework.data.couchbase.core.mapping.Document;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
@@ -73,8 +75,8 @@ public class ReactiveUpsertByIdOperationSupport implements ReactiveUpsertByIdOpe
 				CouchbaseDocument converted = template.support().encodeEntity(o);
 				return template.getCollection(collection).reactive()
 						.upsert(converted.getId(), converted.export(), buildUpsertOptions()).map(result -> {
-							template.support().applyUpdatedCas(object, result.cas());
-							return o;
+							Object updatedObject = template.support().applyUpdatedId(o, converted.getId());
+							return (T) template.support().applyUpdatedCas(updatedObject, result.cas());
 						});
 			}).onErrorMap(throwable -> {
 				if (throwable instanceof RuntimeException) {
--- a/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
@@ -541,6 +541,7 @@ public class MappingCouchbaseConverter extends AbstractCouchbaseConverter implem
 				generatedValueInfo = idProperty.findAnnotation(GeneratedValue.class);
 				String generatedId = generateId(generatedValueInfo, prefixes, suffixes, idAttributes);
 				target.setId(generatedId);
+				// this is not effective if id is Immutable, and accessor.setProperty() returns a new object in getBean()
 				accessor.setProperty(idProperty, generatedId);
 			} else {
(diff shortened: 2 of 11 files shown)


src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -541,6 +541,7 @@ public void doWithPersistentProperty(final CouchbasePersistentProperty prop) {
 				generatedValueInfo = idProperty.findAnnotation(GeneratedValue.class);
 				String generatedId = generateId(generatedValueInfo, prefixes, suffixes, idAttributes);
 				target.setId(generatedId);
+				// this is not effective if id is Immutable, and accessor.setProperty() returns a new object in getBean()
```
mp911de, line 544: Id population should happen on the Template API level, see [MongoDB for reference](https://github.com/spring-projects/spring-data-mongodb/blob/2.0.x/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java#L1054). Note, newer versions have the [Id population extracted](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoTemplate.java#L2717) because it happened in multiple places.

src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
```diff
@@ -71,10 +73,10 @@ public ReactiveUpsertByIdOperationSupport(final ReactiveCouchbaseTemplate templa
 		public Mono<T> one(T object) {
 			return Mono.just(object).flatMap(o -> {
 				CouchbaseDocument converted = template.support().encodeEntity(o);
+				Object updatedObject = template.support().applyUpdatedId(o, converted.getId());
 				return template.getCollection(collection).reactive()
 						.upsert(converted.getId(), converted.export(), buildUpsertOptions()).map(result -> {
-							template.support().applyUpdatedCas(object, result.cas());
```
daschl, line 76: would it make sense to move both apply methods into the upsert map callback so they are done in   one spot?

daschl (changes requested):
just one minor note from me, but looks like mark has another question that probably requires some more changes

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
mikereiche: After adding the update via CouchbaseTemplateSupport.applyUpdatedId() ( called by Reactive[Insert/Upsert]ByIdSupport.one(entity) which is called from the template), I left the update to the sourceObject here for two reasons - the first being that a number of unit tests depend on it (those can be fixed), the second is that if it is removed from here, it would appear to be a regression for anyone calling  `repository.save( sourceObject )`, and expecting sourceObject to contain the id. They should be  calling as  `savedObject = repository.save( sourceObject )`.   I could remove this now in hope that will result in fewer customers using it incorrectly, or wait to do it in the 4.1 release because it could be a breaking change for some.

My first idea was to have the Couchbase implementation of EntityWriter.write(source, target) return the updated object - but the signature returns void. (updating here and returning accessor.getBean() would have worked for that).

src/main/java/org/springframework/data/couchbase/core/ReactiveUpsertByIdOperationSupport.java
mikereiche: For both id and cas updates, they are done as soon as the value is available. 
There is no efficiency benefit to doing them at the same time even for immutable entities - each update will require that a new object be created.

mikereiche: After a lot of thinking, I made this change.

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
mp911de: Let's schedule any removal for the next version. We're in the RC phase with the GA release coming up Oct 28.

In any case in an async/reactive API usage scenario, it's hard to reason about when the model is being updated. In-place updates could have nasty side effects therefore we generally recommend using immutable types if you want to work with the saved/updated object instance that is returned by `save(…)`.

PR #278, opened 2020-11-12
DATACOUCH-588 - Refactoring part 1 of n.

mikereiche:
This is part 1 of a multi-part refactoring effort. 

The files below are the non-test files involved. The classes marked new and old are exactly that, and the files marked 'replace' are replacements for similar classes.  To understand the changes, start with the resolveQuery() of the two repository factory classes, the *PartTeeCouchbaseQuery and *StringBasedCouchbaseQuery that they created, then their execute() method ( AbstractCouchbaseQueryBase.execute() ) which calls the Abstract*CouchbaseQuery.doExecute(), which creates a *CouchbaseQueryExecution and executes it. 

org/springframework/data/couchbase/repository/query/AbstractCouchbaseQuery.java new
org/springframework/data/couchbase/repository/query/AbstractCouchbaseQueryBase.java new
org/springframework/data/couchbase/repository/query/AbstractReactiveCouchbaseQuery.java new
org/springframework/data/couchbase/repository/query/CouchbaseQueryExecution.java new
org/springframework/data/couchbase/repository/query/CouchbaseQueryMethod.java old
org/springframework/data/couchbase/repository/query/DtoInstantiatingConverter.java new
org/springframework/data/couchbase/repository/query/PartTreeCouchbaseQuery.java replace
org/springframework/data/couchbase/repository/query/ReactiveCouchbaseQueryExecution.java new
org/springframework/data/couchbase/repository/query/ReactiveCouchbaseQueryMethod.java new
org/springframework/data/couchbase/repository/query/ReactivePartTreeCouchbaseQuery.java replace
org/springframework/data/couchbase/repository/query/ReactiveStringBasedCouchbaseQuery.java replace
org/springframework/data/couchbase/repository/query/ResultProcessingConverter.java new
org/springframework/data/couchbase/repository/query/StringBasedCouchbaseQuery.java replace
org/springframework/data/couchbase/repository/query/StringN1qlQueryCreator.java old
org/springframework/data/couchbase/repository/support/CouchbaseRepositoryFactory.java old
org/springframework/data/couchbase/repository/support/ReactiveCouchbaseRepositoryFactory.java old


src/test/java/org/springframework/data/couchbase/domain/AirportRepository.java
```diff
@@ -39,6 +42,7 @@
 	Iterable<Airport> findAll();
 
 	@Override
+	@ScanConsistency(query = QueryScanConsistency.REQUEST_PLUS)
 	Airport save(Airport airport);
```
daschl, line 46: isn't that a KV method which does not need the scan consistency?

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -136,13 +141,19 @@ void findBySimpleProperty() {
 		Airport vie = null;
 		try {
 			vie = new Airport("airports::vie", "vie", "loww");
-			airportRepository.save(vie);
+			vie = airportRepository.save(vie);
 			List<Airport> airports = airportRepository.findAllByIata("vie");
-			assertEquals(vie.getId(), airports.get(0).getId());
+			assertEquals(1, airports.size());
+			System.out.println("findAllByIata(0): " + airports.get(0));
```
daschl, line 147: are these printlns leftovers from debugging in the tests?

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -93,9 +102,19 @@ void findBySimpleProperty() {
...
 			List<Airport> airports1 = airportRepository.findAllByIata("vie").collectList().block();
-			assertEquals(1,airports1.size());
+			assertEquals(1, airports1.size());
 			List<Airport> airports2 = airportRepository.findAllByIata("vie").collectList().block();
-			assertEquals(1,airports2.size());
+			assertEquals(1, airports2.size());
+			vie = airportRepository.save(vie).block();
+			List<Airport> airports = airportRepository.findAllByIata("vie").collectList().block();
+			assertEquals(1, airports.size());
+			System.out.println("findAllByIata(0): " + airports.get(0));
```
daschl, line 111: looks like some debugging leftovers - I'd recommend to remove all those prints from tests since they make them very noisy in logfiles

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -124,6 +143,23 @@ void count() {
 				airportRepository.save(airport).block();
 			}
 
+			int page = 0;
+			for (; page < 10;) {
+				PageRequest pageable = PageRequest.of(page++, 2);
+				System.out.println("findAllByIataLike: " + "%S");
+				// java.lang.IndexOutOfBoundsException: Source emitted more than one item
+				// Flux<Airport> airportFlux = airportRepository.findAllByIataLike("S%");
```
daschl, line 151: can this be removed or does this needs to be handled?

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -124,6 +143,23 @@ void count() {
 				airportRepository.save(airport).block();
 			}
 
+			int page = 0;
+			for (; page < 10;) {
```
daschl, line 147: a while(page < 10) is probably better here

src/main/java/org/springframework/data/couchbase/repository/query/PartTreeCouchbaseQuery.java
```diff
@@ -0,0 +1,138 @@
...
+	 * @param evaluationContextProvider must not be {@literal null}.
+	 */
+	public PartTreeCouchbaseQuery(CouchbaseQueryMethod method, CouchbaseOperations operations,
+			SpelExpressionParser expressionParser, QueryMethodEvaluationContextProvider evaluationContextProvider) {
+
+		super(method, operations, expressionParser, evaluationContextProvider);
+
+		this.processor = method.getResultProcessor();
+		// @dashl - I don't know the purpose of this for(), it was in the code I copied
+		for (PartTree.OrPart parts : this.tree = new PartTree(method.getName(),
```
daschl, line 61: it looks like this isn't doing anything? probably some leftover - if you think it can be removed, please do

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedCouchbaseQuery.java
```diff
@@ -0,0 +1,103 @@
...
+		if (LOG.isDebugEnabled()) {
+			LOG.debug("Created query "+ query.export());
+		}
+
+		return query;
+	}
+
+	@Override
+	protected Query createCountQuery(ParametersParameterAccessor accessor) {
+		return /*applyQueryMetaAttributesWhenPresent*/(createQuery(accessor)); // not yet implemented
```
daschl, line 91: should this throw an exception for now, or can we delegate to somewhere? if we leave it like this, would the refactor reduce the features we had (breaking code?)

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedCouchbaseQuery.java
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	/*
+	 * (non-Javadoc)
+	 * @see org.springframework.data.couchbase.repository.query.AbstractCouchbaseQuery#isLimiting()
+	 */
+	@Override
+	protected boolean isLimiting() {
+		return false; // not yet implemented
```
daschl, line 100: same here, as above

src/main/java/org/springframework/data/couchbase/repository/query/ResultProcessingConverter.java
```diff
@@ -0,0 +1,74 @@
+package org.springframework.data.couchbase.repository.query;
```
daschl, line 1: license header missing

src/main/java/org/springframework/data/couchbase/repository/query/ReactiveStringBasedCouchbaseQuery.java
```diff
@@ -0,0 +1,106 @@
...
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+import org.springframework.util.Assert;
+
+/**
+ * Query to use a plain JSON String to create the {@link Query} to actually execute.
+ *
+ * @author Michael Reiche
+ * @since 4.1
+ */
+public class ReactiveStringBasedCouchbaseQuery extends AbstractReactiveCouchbaseQuery {
```
daschl, line 35: this class looks very similar to the non-reactive one (same comments apply). Could some of the methods be refactored into a parent that extends the AbstractReactiveCouchbaseQuery (so a common string based one that has the common logic)

src/main/java/org/springframework/data/couchbase/repository/query/ReactivePartTreeCouchbaseQuery.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.expression.spel.standard.SpelExpressionParser;
+
+/**
+ * Reactive PartTree {@link RepositoryQuery} implementation for Couchbase.
+ * Replaces ReactivePartN1qlBasedQuery
+ *
+ * @author Michael Reiche
+ * @since 4.1 
+ */
+public class ReactivePartTreeCouchbaseQuery extends AbstractReactiveCouchbaseQuery {
```
daschl, line 38: Same comment as with the string equivalent, is there a common base class for reactive and non-reactive that encapsulates the commonalities maybe?

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryMethod.java
```diff
@@ -182,6 +185,24 @@ public String getInlineN1qlQuery() {
 		return StringUtils.hasText(query) ? query : null;
 	}
 
+	/**
+	 * is this a 'delete'?
+	 *
+	 * @return is this a 'delete'?
+	 */
+	public boolean isDeleteQuery() {
+		return getName().toLowerCase().startsWith("delete");
```
daschl, line 194: should we use Locale.ROOT in the toLowerCase?

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryMethod.java
```diff
@@ -182,6 +185,24 @@ public String getInlineN1qlQuery() {
...
+	}
+
+	/**
+	 * is this an 'exists' query?
+	 *
+	 * @return is this an 'exists' query?
+	 */
+	public boolean isExistsQuery() {
+		return getName().toLowerCase().startsWith("exists");
```
daschl, line 203: should we use Locale.ROOT in the toLowerCase?

src/main/java/org/springframework/data/couchbase/repository/query/ReactiveCouchbaseQueryExecution.java
```diff
@@ -0,0 +1,85 @@
...
+			this.method = method;
+		}
+
+		/*
+		 * (non-Javadoc)
+		 * @see org.springframework.data.couchbase.repository.query.AbstractCouchbaseQuery.Execution#execute(org.springframework.data.couchbase.core.query.Query, java.lang.Class, java.lang.String)
+		 */
+		@Override
+		public Object execute(Query query, Class<?> type, String collection) {
+			return operations.removeByQuery(type)/*.inCollection(collection)*/.matching(query).all();
```
daschl, line 58: does the collection string needs to be handled?

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryExecution.java
```diff
@@ -0,0 +1,157 @@
...
+			this.method = method;
+		}
+
+		/*
+		 * (non-Javadoc)
+		 * @see org.springframework.data.couchbase.repository.query.AbstractCouchbaseQuery.Execution#execute(org.springframework.data.couchbase.core.query.Query, java.lang.Class, java.lang.String)
+		 */
+		@Override
+		public Object execute(Query query, Class<?> type, String collection) {
+			return operations.removeByQuery(type).matching(query).all();
```
daschl, line 63: the collection string here is not handled at all - should we fail explicitly if it is present or even better handle it?

daschl (changes requested):
Couple remarks. Also note that there are many "not yet implement" comments in the code - I'm okay with these as long as they are not degrading functionality we had before. If so, we should probably work on this in a feature branch until we are back to feature parity with the stable code, so we don't break folks.

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
mikereiche: removed

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
mikereiche: removed

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
mikereiche: I've flattened out the loop to better verify the results.

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryMethod.java
mikereiche: done on exists, delete and count.  Mark suggested an annotation.  I will (eventually) see what mongo does.

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryMethod.java
mikereiche: done

src/main/java/org/springframework/data/couchbase/repository/query/PartTreeCouchbaseQuery.java
mikereiche: removed

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedCouchbaseQuery.java
mikereiche: I've added the implementation from Couchbase. It handles a "Meta" annotation (which will be ignored by Couchbase for now).

src/main/java/org/springframework/data/couchbase/repository/query/ResultProcessingConverter.java
mikereiche: fixed.

src/main/java/org/springframework/data/couchbase/repository/query/ReactiveCouchbaseQueryExecution.java
mikereiche: The do eventually.  The support of an inCollection(collection) was one of the things I backed out to keep the (initial) changes to a minium.

src/main/java/org/springframework/data/couchbase/repository/query/ReactiveStringBasedCouchbaseQuery.java
mikereiche: The two classes are identical, other than one extends AbstractCouchbaseQuery and the other extends AbstractReactiveCouchbaseQuery - and the first two args of the constructor are the regular and reactive versions.  However, I don't see a clean way to combine the common methods while keeping the inherited classes different.

src/test/java/org/springframework/data/couchbase/domain/AirportRepository.java
mikereiche: fixed

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
mikereiche: I changed this to actually use the paging. Removed the commented-out lines and added assertions.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedCouchbaseQuery.java
mikereiche: I left this as-is for now.  We don't have an annotation or method-name-scheme to determine if a query contains "limit".  And we don't do anything different if it is.

src/main/java/org/springframework/data/couchbase/repository/query/ReactivePartTreeCouchbaseQuery.java
mikereiche: It seems there should be a way to make this tidier, but it eludes me.

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryExecution.java
mikereiche: As above - we are going to handle it, but I wanted to keep that out of the initial branch.

pom.xml
```diff
@@ -254,6 +261,14 @@
                 <groupId>org.asciidoctor</groupId>
                 <artifactId>asciidoctor-maven-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>9</source>
```
daschl, line 268: We are targeting java 8, not 9 right?

daschl: Also, I would think this is part of the parent pom?

src/main/java/org/springframework/data/couchbase/core/query/Meta.java
```diff
@@ -0,0 +1,306 @@
...
+    return ObjectUtils.nullSafeEquals(this.flags, other.flags);
+  }
+
+  /**
+   * {@link CursorOption} represents {@code OP_QUERY} wire protocol flags to change the behavior of queries.
+   *
+   * @author Christoph Strobl
+   * @since 1.10
+   */
+  public enum CursorOption {
```
daschl, line 275: we do not have cursors, does this apply at all to the impl?

src/main/java/org/springframework/data/couchbase/repository/Meta.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Sets the number of documents to return per batch. <br />
+   * Use {@literal 0 (zero)} for no limit. A <strong>negative limit</strong> closes the cursor after returning a single
+   * batch indicating to the server that the client will not ask for a subsequent one.
+   *
+   * @return {@literal 0 (zero)} by default.
+   * @since 2.1
+   */
+  int cursorBatchSize() default 0;
```
daschl, line 52: there are many references to cursors, comments etc which are all very mongo-ish but do not apply to couchbase

src/main/java/org/springframework/data/couchbase/repository/query/CouchbaseQueryMethod.java
```diff
@@ -171,6 +178,60 @@ public ScanConsistency getScanConsistencyAnnotation() {
...
+			return new org.springframework.data.couchbase.core.query.Meta();
+		}
+
+		org.springframework.data.couchbase.core.query.Meta metaAttributes = new org.springframework.data.couchbase.core.query.Meta();
+		if (meta.maxExecutionTimeMs() > 0) {
+			metaAttributes.setMaxTimeMsec(meta.maxExecutionTimeMs());
+		}
+
+		if (meta.cursorBatchSize() != 0) {
+			metaAttributes.setCursorBatchSize(meta.cursorBatchSize());
```
daschl, line 214: does not apply? (see other comments)

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/repository/query/ReactiveStringBasedCouchbaseQuery.java
mikereiche: Maybe interfaces with default method implementations?

pom.xml
mikereiche: I vaguely recall adding this when prompted by intellij to facilate some <class>.of(...) call.  Will check

mikereiche: removed. It was for Set.of(...) in a test.

src/main/java/org/springframework/data/couchbase/repository/Meta.java
mikereiche: I removed cursor, comment etc.  I left maxExecutionTimeMs just so Meta isn't empty.

src/main/java/org/springframework/data/couchbase/core/query/Meta.java
```diff
@@ -0,0 +1,306 @@
...
+/**
+ * Meta-data for {@link Query} instances.
+ *
+ * @author Michael Reiche
+
+ */
+public class Meta {
+
+  private enum MetaKey {
+    MAX_TIME_MS("$maxTimeMS"), MAX_SCAN("$maxScan"), COMMENT("$comment"), SNAPSHOT("$snapshot");
```
daschl, line 41: this stuff is mongo related, right? why do we keep this?

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/core/query/Meta.java
mikereiche: I missed this when I cleaned out the Meta annotation. I've removed all the mongo-specific elements, I've left just an EXAMPLE in the enum because it can't be empty.  I've also remove the mongo maxTimeMs from the Meta annotation.

daschl (approved):

PR #281, opened 2020-12-08
DATACOUCH-661 - Fix integrations tests from 650

mikereiche:
The problems were in the tests themselves.  The changes in datacouch-650 are fine.
These only run with "mvn integration-test", so the failures were not being caught.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
@@ -16,7 +16,22 @@
 
 package org.springframework.data.couchbase.repository;
 
-import com.couchbase.client.core.error.IndexExistsException;
+import static java.util.Arrays.asList;
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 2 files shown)


src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -128,8 +130,11 @@ void count() {
 				Airport airport = new Airport("airports::" + iatas[i], iatas[i] /*iata*/, iatas[i].toLowerCase() /* lcao */);
 				airportRepository.save(airport).block();
 			}
-
-			Long airportCount = airportCount = airportRepository.count().block();
+			List<Airport> airports = airportRepository.findAll().collectList().block();
+			for (Airport ap : airports) {
+				System.out.println(ap);
```
daschl, line 135: is the println needed?

mikereiche: Not anymore.  There were 8 airports after inserting 7.  The println was to find the extra one - it was was because the deleteAll() in an earlier test did not block.  I'll remove the whole findAll()/for

PR #285, opened 2020-12-08
DATACOUCH-630 Add expiry option to replace.

mikereiche:
Add expiry option to 'replace' similar to how it is on 'insert' and 'update'

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
@@ -18,9 +18,11 @@ package org.springframework.data.couchbase.core;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
+import java.time.Duration;
 import java.util.Collection;
 
 import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
+import org.springframework.data.couchbase.core.mapping.Document;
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
```diff
@@ -93,6 +97,13 @@ private ReplaceOptions buildReplaceOptions(T object) {
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
+			if (expiry != null && !expiry.isZero()) {
+				options.expiry(expiry);
+			} else if (domainType.isAnnotationPresent(Document.class)) {
+				Document documentAnn = domainType.getAnnotation(Document.class);
+				long durationSeconds = documentAnn.expiryUnit().toSeconds(documentAnn.expiry());
```
daschl, line 104: could some of these values (like the expiry unit be null and cause a NPE if not set?)

daschl (approved):
+2, but with a comment on a potential NPE . if that is a no-issue feel free to merge

src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java
mikereiche: Document annotation has a default expiryUnit of SECONDS and a default expiry of 0

PR #292, opened 2020-12-11
DATACOUCH-645 - Support Document(expiryExpression) annotation.

mikereiche:
1) get the expiryExpression from the converted CouchbaseDocument.

2) added interfaces for apis on core objects which are common among
many objects. For instance withExpiry and withExpirty on all the
insert/replace/upsert/remove objects.  The definition of an interface
simplyfies testing of all the possible combinations - instead of having
four (or eight, counting Reactive), there is one interface to test
them all.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/InCollection.java
+++ b/src/main/java/org/springframework/data/couchbase/core/InCollection.java
@@ -13,21 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-package org.springframework.data.couchbase.domain;
-
-import org.springframework.data.couchbase.core.mapping.Document;
+package org.springframework.data.couchbase.core;
 
 /**
- * Annotated User entity for tests
+ * A common interface for all of Insert, Replace, Upsert that take collection
  *
  * @author Michael Reiche
+ *
+ * @param <T> - the entity class
  */
-
-@Document(expiry = 1)
-public class UserAnnotated extends User {
-
-	public UserAnnotated(String id, String firstname, String lastname) {
-		super(id, firstname, lastname);
-	}
+public interface InCollection<T> {
+ Object inCollection(String collection);
 }
--- a/src/main/java/org/springframework/data/couchbase/core/WithExpiry.java
+++ b/src/main/java/org/springframework/data/couchbase/core/WithExpiry.java
@@ -13,21 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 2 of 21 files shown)


src/main/java/org/springframework/data/couchbase/core/InCollection.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * A common interface for all of Insert, Replace, Upsert that take collection
+ *
+ * @author Michael Reiche
+ *
+ * @param <T> - the entity class
+ */
+public interface InCollection<T> {
+ Object inCollection(String collection);
```
mp911de, line 26: Mind the formatting. What's the reason for extracting all these interfaces into top-level ones? From a usage perspective, calling code would not mix replace and e.g. remove operations by reusing the same interface, would it?

src/main/java/org/springframework/data/couchbase/core/WithExpiry.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A common interface for all of Insert, Replace, Upsert that take expiry
+ *
+ * @author Michael Reiche
+ *
+ * @param <T> - the entity class
+ */
+public interface WithExpiry<T> {
+  Object withExpiry(Duration expiry);
```
mp911de, line 28: How would you express that a default expiry should be overridden to persistent (i.e. no expiry)?

src/main/java/org/springframework/data/couchbase/core/InCollection.java
mikereiche: fixed the formatting in this and the other interfaces.
From a usage perspective, the interfaces are not helpful.  However, from a testing perspective, they allow testing all the implementers of those interfaces more easily.   For testing withExpiry(), @Document( expiry), and @Document(@expiryExpression) (which require three different entity classes) across insert, replace and upsert, that would take 3 X 3 = 9 test cases. By having insert, replace and upsert implement OneAndAll and WithExpiry, all these tests can be done in a loop as in withExpiryAndExpiryAnnotation() 
https://github.com/spring-projects/spring-data-couchbase/blob/7815522d682fceb0f55082ff825e6dabbd8a9343/src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateKeyValueIntegrationTests.java
(I was not able to figure out how to define an interface that would be also satisfied by the one() and all() of Remove)
I did promise @daschl that I would split this into two branches - one for the interfaces followed by a second for expiry - but it's hard to explain the interfaces without showing how they would be used.

PR #294, opened 2020-12-14
DATACOUCH-652 - Inject translation service bean in CouchbaseTemplateSupport.

jorgerod:
[DATACOUCH-652](https://jira.spring.io/projects/DATACOUCH/issues/DATACOUCH-652)

TranslationService is defined as bean but in CouchbaseTemplateSupport. 
That bean is not being used and a new instance of JacksonTranslationSerivce is being created in the constructor.

We want CouchbaseTemplateSupport to receive a TranslationService type bean.

- [X] You have read the [Spring Data contribution guidelines](https://github.com/spring-projects/spring-data-build/blob/master/CONTRIBUTING.adoc).
- [X] There is a ticket in the bug tracker for the project in our [JIRA](https://jira.spring.io/browse/DATACOUCH).

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -62,6 +62,7 @@ import com.couchbase.client.java.json.JacksonTransformers;
  * @author Simon Baslé
  * @author Stephane Nicoll
  * @author Subhashni Balakrishnan
+ * @author Jorge Rodriguez Martin
  */
 @Configuration
 public abstract class AbstractCouchbaseConfiguration {
@@ -146,14 +147,26 @@ public abstract class AbstractCouchbaseConfiguration {
 
 	@Bean(name = BeanNames.COUCHBASE_TEMPLATE)
 	public CouchbaseTemplate couchbaseTemplate(CouchbaseClientFactory couchbaseClientFactory,
+			MappingCouchbaseConverter mappingCouchbaseConverter, TranslationService couchbaseTranslationService) {
+		return new CouchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter, couchbaseTranslationService);
+	}
+
+	public CouchbaseTemplate couchbaseTemplate(CouchbaseClientFactory couchbaseClientFactory,
 			MappingCouchbaseConverter mappingCouchbaseConverter) {
-		return new CouchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter);
+		return couchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter, new JacksonTranslationService());
 	}
 
 	@Bean(name = BeanNames.REACTIVE_COUCHBASE_TEMPLATE)
 	public ReactiveCouchbaseTemplate reactiveCouchbaseTemplate(CouchbaseClientFactory couchbaseClientFactory,
+			MappingCouchbaseConverter mappingCouchbaseConverter, TranslationService couchbaseTranslationService) {
+		return new ReactiveCouchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter,
+				couchbaseTranslationService);
+	}
+
+	public ReactiveCouchbaseTemplate reactiveCouchbaseTemplate(CouchbaseClientFactory couchbaseClientFactory,
 			MappingCouchbaseConverter mappingCouchbaseConverter) {
-		return new ReactiveCouchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter);
+		return reactiveCouchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter,
+				new JacksonTranslationService());
 	}
 
 	@Bean(name = BeanNames.COUCHBASE_OPERATIONS_MAPPING)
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@jorgerod Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=294)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=294) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jorgerod Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-data-couchbase&pullRequestId=294)!

mikereiche:
This looks good to me. I'm adding another reviewer.

jorgerod:
Hi @mikereiche @daschl 

I think it would be interesting to make a backport to 4.1.x.

How do you see it?

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -145,14 +146,14 @@ protected void configureEnvironment(final ClusterEnvironment.Builder builder) {
 
 	@Bean(name = BeanNames.COUCHBASE_TEMPLATE)
 	public CouchbaseTemplate couchbaseTemplate(CouchbaseClientFactory couchbaseClientFactory,
-			MappingCouchbaseConverter mappingCouchbaseConverter) {
-		return new CouchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter);
+			MappingCouchbaseConverter mappingCouchbaseConverter, TranslationService translationService) {
```
daschl, line 149: this breaks backwards compatibility by introducing another required argument - if this should be added we need another overload that keeps the defaults as is I think

daschl (changes requested):
I think this contains a breaking change as-is

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
jorgerod: Hi @daschl 

You are right about the overloading of the _CouchbaseTemplate_ and _ReactiveCouchbaseTemplate_ constructors. I'll update it now.

As for the definition of beans _couchbaseTemplate_ and _reactiveCouchbaseTemplate_, do you consider any change necessary? I didn't quite understand you.

daschl: Yeah, I was actually also talking about the bean definitions - because if someone overloads them in their actual configuration, just changing the signature on update will fail to compile for them.

mikereiche, line 149: The translationService argument should be named couchbaseTranslationService to match the bean name.
When the argument name does not match the bean name, the spring framework can only match by the type, and if there is ever another bean of type TranslationService added, the spring framework will not be able to automatically match this argument with the couchbaseTranslationService bean.

jorgerod: Hi @daschl 

It's true that is a breaking change, but I can't think of any other way to do it. Bean _couchbaseTranslateService_ is being created but then not used.


If you don't want to be breakers, we could keep the overload on the _CouchbaseTemplate_ and _ReactiveCouchbaseTemplate_ constructors and at another time add the TranslationService parameter to the configuration beans and then be breakers.

What do you think?

Thank you very much

daschl: @jorgerod well, you could make the beans take the new parameter by default (as you do now), but add another 2 methods that look like the old signature and not be a Bean, which call into the new beans and just initialize the json translation service. Then if someone overrides it, their code would not break but everyone can then switch over to the new bean if needed?

jorgerod: I understand. Thank you very much for the help  @daschl

mikereiche:
@jorgerod - Please put a period at the end of the first line of the commit message. (they are parsed programmatically) Thanks.

DATACOUCH-652 - Inject translation service bean in CouchbaseTemplateSupport

jorgerod:
> @jorgerod - Please put a period at the end of the first line of the commit message. (they are parsed programmatically) Thanks.
> 
> [DATACOUCH-652](https://jira.spring.io/browse/DATACOUCH-652) - Inject translation service bean in CouchbaseTemplateSupport

Hi @mikereiche 

I've checked the commit messages and they all have a period at the end of the line.

jorgerod:
Hello @daschl @mikereiche 

Have you been able to see the latest updates?

Greetings

jorgerod:
Hello @daschl @mikereiche 

Have you been able to see the latest updates?

Greetings

mikereiche:
@jorgerod - can you resolve the conflicts? Then I will merge this.

jorgerod:
@mikereiche Conflicts resolved.

mp911de:
@jorgerod please squash your commits into a single one as we keep a single commit (and a linear history) per change.

jorgerod:
Hi @mp911de 

I have already squashes my commits. Does it look good?

Thank you

jorgerod:
> Hi @mikereiche @daschl
> 
> I think it would be interesting to make a backport to 4.1.x.
> 
> How do you see it?

Hi @mikereiche 

How does it look?

mikereiche:
It will be in the next 4.1.x release scheduled for February 10.

PR #295, opened 2020-12-15
Datacouch 550 autoindex not working in spring boot app

mikereiche:
This needs some discussion as to whether or not we want to support this.
I created the branch to capture the debugging and changes I had made
earlier when Denis asked me to investigate this issue.
Mark Paluch has indicated this is maybe not a good idea.

Due to the interdependencies of beans, autoindexing does not work
in the startup of a normal spring-boot application. Processing of
entities occurs during the initialization of CouchbaseMappinContext,
which occurs before initialization of MappingCouchbaseConverter,
which occurs before the initialization of CouchbaseTemplate -
which creates the indexCreator listener. So when the
AbstractMappingContext (CouchbaseMappingContext) is publishing
MappingContextEvents - there is not yet any indexCreator.
And subsequent processing of entities finds the entities cached,
and therefore does not publish MappingContextEvents. This change
overrides the addPersistentEntity() and getPersistentEntity()
methods of AbstractMappingContext such that caching does not
preven the MappingContextEvents from being published.  This change
also exposes indexCreator classesSeen map to CouchbaseMappingContext
so that once an entity has been processed by indexCreator, it is
not published again.

Autoindexing by SpringJUnitConfig did work prior to this change.

There is no test case as our automated testing does not create a spring-boot app.
- [-] You submit test cases (unit or integration tests) that back your changes.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
@@ -166,6 +166,10 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 
 		if (context instanceof ConfigurableApplicationContext && indexCreator != null) {
 			((ConfigurableApplicationContext) context).addApplicationListener(indexCreator);
+			if (mappingContext instanceof CouchbaseMappingContext) {
+				CouchbaseMappingContext cmc = (CouchbaseMappingContext) mappingContext;
+				cmc.setIndexCreator(indexCreator);
+			}
 		}
 	}
 }
--- a/src/main/java/org/springframework/data/couchbase/core/index/CouchbasePersistentEntityIndexCreator.java
+++ b/src/main/java/org/springframework/data/couchbase/core/index/CouchbasePersistentEntityIndexCreator.java
@@ -121,4 +121,7 @@ public class CouchbasePersistentEntityIndexCreator implements ApplicationListene
 		return this.mappingContext.equals(context);
 	}
 
+	public boolean hasSeen(CouchbasePersistentEntity<?> entity) {
+		return classesSeen.containsKey(entity.getType());
+	}
 }
--- a/src/main/java/org/springframework/data/couchbase/core/mapping/CouchbaseMappingContext.java
+++ b/src/main/java/org/springframework/data/couchbase/core/mapping/CouchbaseMappingContext.java
@@ -16,10 +16,13 @@
 
 package org.springframework.data.couchbase.core.mapping;
 
+import java.util.Optional;
+
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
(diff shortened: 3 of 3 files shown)


src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -166,6 +166,10 @@ private void prepareIndexCreator(final ApplicationContext context) {
 
 		if (context instanceof ConfigurableApplicationContext && indexCreator != null) {
 			((ConfigurableApplicationContext) context).addApplicationListener(indexCreator);
+			if (mappingContext instanceof CouchbaseMappingContext) {
```
aaronjwhiteside, line 169: would it not make sense to take this opportunity to move the glueing of beans out of the main code path and into a @Configuration class?

aaronjwhiteside:
This might solve the issue, but I think it's adding more complexity into the mix.. 

I think it might be cleaner if `CouchbasePersistentEntityIndexCreator` creation was moved completely out of the `CouchbaseTemplate` and into a `@Configuration` class.

I imagine it would take `CouchbaseOperations` and `MappingContext` as constructor arguments, and implement `InitializingBean`. 

In `afterPropertiesSet()` you could simply call `MappingContext#getPersistentEntities()` and iterate over the results doing what it currently does.

No need to listen to any events, and no need to implement complex interactions/callbacks with `ApplicationContextAware`.

mikereiche:
@aaronjwhiteside 
"This might solve the issue, but I think it's adding more complexity into the mix.."
I don't have the background of why it was implemented the way it was - with events published to an application listener - I can only assume there were reasons - perhaps because autoindexing could take several minutes and holding up application initialization for that long is not acceptable (although the alternative of having the application initialized before the indexes are generated and that creates its own issues).  Or perhaps because that is the point at which all persistent entities are processed.  Probably both. 
 I just fixed the mechanism that was already there by ensuring that events would be sent to the listener after it was listening - which was the original intent.

aaronjwhiteside:
@mikereiche events can continue to be produced as they are now, but you do not have to use the events to trigger the automate creation of the indices.

If something needs to be async, then that should be explicit, i.e. an Executor in the `CouchbasePersistentEntityIndexCreator`

But even that doesn't hold weight for me, like you said the indices should be available prior to the application starting.

And I'm just remember my own trouble with trying to parallelize index creation in Couchbase, the short answer is it's not supported.

Should we not strive to make things as simple as they can be, but no simpler? Right now we're adding workarounds for a bad design..

mikereiche:
Make a pull request.

mikereiche:
> trying to parallelize index creation in Couchbase, the short answer is it's not supported.

the onApplicationEvent() calls are made serially. That's not an issue.

aaronjwhiteside:
@mikereiche please see https://github.com/spring-projects/spring-data-couchbase/pull/298

PR #1040, opened 2021-01-07
DATACOUCH-588 - Part 2 of framework changes. Add support for projecti…

mikereiche:
…on and distinct.

Support for projection is only for properties of the top-level entity. For
instance, in UserSubmission, only the properties below can be specified in
 the projection. Projection support does not provide means of specifying
something like address.street - you can only project (or not project) the
whole address property. However, the address type in your resultType could
have a subset of the properties in Address.

If the corresponding submissions in the resultType
contained only the userId property

public class UserSubmission extends ComparableEntity {
	private String id;
	private String username;
	private List<String> roles;
	private Address address;
	private List<Submission> submissions;

Support for Distinct - I have appropriated the MongoDB model for Distinct.
It defines a separate DistinctOperationSupport class (within
ExecutableFindByQuerySupport) which supports the distinct( distinctFields )
api and execution. The DistinctOperationSupport class has only a
distinctFields member, and a 'delegate' member, which is an
ExecutableFindByQuerySupport object. TBH, I don't see the advantage
over simply adding a distinctFields member to ExecutableFindByQuerySupport

Amend #1 - changes as discussed in Pull Request
         - clean up test entity types

Amend #2
- Eliminate DistinctOperationSupport class. In MongoDB, only distinct on a single field is supported, so the returnType from distinct was very different from the returnType of other query operations (all(), one() etc. (but so is count(), and it doesn't need it's own class)). In Couchbase, distinct on any fields in the entity is allowed - so the returned type could be the domainType or resultType. And as(resultType) still allows any resultType to be specified. This makes it unnecessary to have combinations of interfaces such as DistinctWithProjection and DistinctWithQuery.

- Clean up the interfaces in ExecutableFindByQuery. There are two types of interfaces (a) the TerminatingFindByQuery which has the one(), oneValue() first(), firstValue(), all(), count(), exists() and stream(); and (b) the option interfaces (FindByQueryWithConsistency etc), which are essentially with-er interfaces. The changes are:
1) make all the with-er interfaces base interfaces instead of chaining them together. (I don't know why there isn't simply one interface with all the with-er methods).
2) make the ExecutableFindByQuery interface extend the Terminating interface and all the with-er interfaces.

Amend #3
- Add execution support for collections

Amend #4
- Add tests for collections. This includes a new CollectionAwareIntegrationTests class which extends a new JavaIntegratationTests class which extends the existing ClusterAwareIntegrationTests.
- Fixed up several issues collections issues that were uncovered by the tests.
- Did further cleanup of OperationSupport interfaces.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperation.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperation.java
@@ -19,11 +19,14 @@ import java.util.List;
 import java.util.Optional;
 import java.util.stream.Stream;
 
-import com.couchbase.client.java.analytics.AnalyticsScanConsistency;
-import com.couchbase.client.java.query.QueryScanConsistency;
 import org.springframework.dao.IncorrectResultSizeDataAccessException;
 import org.springframework.data.couchbase.core.query.AnalyticsQuery;
+import org.springframework.data.couchbase.core.support.OneAndAll;
+import org.springframework.data.couchbase.core.support.WithAnalyticsConsistency;
+import org.springframework.data.couchbase.core.support.WithAnalyticsQuery;
 import org.springframework.lang.Nullable;
+
+import com.couchbase.client.java.analytics.AnalyticsScanConsistency;
 
 public interface ExecutableFindByAnalyticsOperation {
 
@@ -34,7 +37,7 @@ public interface ExecutableFindByAnalyticsOperation {
 	 */
 	<T> ExecutableFindByAnalytics<T> findByAnalytics(Class<T> domainType);
 
-	interface TerminatingFindByAnalytics<T> {
+	interface TerminatingFindByAnalytics<T> extends OneAndAll<T> {
 
 		/**
 		 * Get exactly zero or one result.
@@ -102,7 +105,7 @@ public interface ExecutableFindByAnalyticsOperation {
 
 	}
 
-	interface FindByAnalyticsWithQuery<T> extends TerminatingFindByAnalytics<T> {
+	interface FindByAnalyticsWithQuery<T> extends TerminatingFindByAnalytics<T>, WithAnalyticsQuery<T> {
 
 		/**
 		 * Set the filter for the analytics query to be used.
@@ -114,6 +117,7 @@ public interface ExecutableFindByAnalyticsOperation {
 
 	}
 
+	@Deprecated
 	interface FindByAnalyticsConsistentWith<T> extends FindByAnalyticsWithQuery<T> {
 
 		/**
@@ -121,10 +125,22 @@ public interface ExecutableFindByAnalyticsOperation {
 		 *
 		 * @param scanConsistency the custom scan consistency to use for this analytics query.
 		 */
+		@Deprecated
 		FindByAnalyticsWithQuery<T> consistentWith(AnalyticsScanConsistency scanConsistency);
 
 	}
 
-	interface ExecutableFindByAnalytics<T> extends FindByAnalyticsConsistentWith<T> {}
+	interface FindByAnalyticsWithConsistency<T> extends FindByAnalyticsConsistentWith<T>, WithAnalyticsConsistency<T> {
+
+		/**
+		 * Allows to override the default scan consistency.
+		 *
+		 * @param scanConsistency the custom scan consistency to use for this analytics query.
+		 */
+		FindByAnalyticsConsistentWith<T> withConsistency(AnalyticsScanConsistency scanConsistency);
+
+	}
+
+	interface ExecutableFindByAnalytics<T> extends FindByAnalyticsWithConsistency<T> {}
 
 }
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperation.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperation.java
@@ -17,6 +17,10 @@ package org.springframework.data.couchbase.core;
 
 import java.util.Collection;
 
+import org.springframework.data.couchbase.core.support.OneAndAllId;
+import org.springframework.data.couchbase.core.support.WithCollection;
+import org.springframework.data.couchbase.core.support.WithProjectionId;
+
 public interface ExecutableFindByIdOperation {
 
 	/**
@@ -26,7 +30,7 @@ public interface ExecutableFindByIdOperation {
 	 */
 	<T> ExecutableFindById<T> findById(Class<T> domainType);
 
-	interface TerminatingFindById<T> {
+	interface TerminatingFindById<T> extends OneAndAllId<T> {
 
 		/**
 		 * Finds one document based on the given ID.
@@ -46,7 +50,7 @@ public interface ExecutableFindByIdOperation {
 
 	}
 
-	interface FindByIdWithCollection<T> extends TerminatingFindById<T> {
+	interface FindByIdWithCollection<T> extends TerminatingFindById<T>, WithCollection<T> {
 
 		/**
 		 * Allows to specify a different collection than the default one configured.
@@ -57,7 +61,7 @@ public interface ExecutableFindByIdOperation {
 
 	}
 
-	interface FindByIdWithProjection<T> extends FindByIdWithCollection<T> {
+	interface FindByIdWithProjection<T> extends FindByIdWithCollection<T>, WithProjectionId<T> {
 
 		/**
 		 * Load only certain fields for the document.
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperationSupport.java
@@ -15,12 +15,11 @@
  */
 package org.springframework.data.couchbase.core;
 
-import org.springframework.data.couchbase.core.ReactiveFindByIdOperationSupport.ReactiveFindByIdSupport;
-
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 
+import org.springframework.data.couchbase.core.ReactiveFindByIdOperationSupport.ReactiveFindByIdSupport;
 import org.springframework.util.Assert;
 
 public class ExecutableFindByIdOperationSupport implements ExecutableFindByIdOperation {
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByQueryOperation.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByQueryOperation.java
@@ -21,6 +21,13 @@ import java.util.stream.Stream;
 
 import org.springframework.dao.IncorrectResultSizeDataAccessException;
 import org.springframework.data.couchbase.core.query.Query;
+import org.springframework.data.couchbase.core.query.QueryCriteriaDefinition;
+import org.springframework.data.couchbase.core.support.OneAndAll;
+import org.springframework.data.couchbase.core.support.WithCollection;
+import org.springframework.data.couchbase.core.support.WithConsistency;
+import org.springframework.data.couchbase.core.support.WithDistinct;
+import org.springframework.data.couchbase.core.support.WithProjection;
+import org.springframework.data.couchbase.core.support.WithQuery;
 import org.springframework.lang.Nullable;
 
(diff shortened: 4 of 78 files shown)


src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperation.java
```diff
@@ -140,55 +140,36 @@
...
 	}
 
 	/**
 	 * Distinct Find support.
 	 *
-	 * @author Christoph Strobl
-	 * @since 2.1
+	 * @author Michael Reiche
 	 */
 	interface FindDistinct {
```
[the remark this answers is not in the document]
daschl: so do you suggest we make those changes? I guess since it is additive it won't be breaking anyone?

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperation.java
```diff
@@ -215,6 +196,13 @@
 		 * @since 3.0
 		 */
 		TerminatingDistinct<T> matching(Query query);
+
+		/**
+		 * Allows to override the default scan consistency.
+		 *
+		 * @param scanConsistency the custom scan consistency to use for this query.
+		 */
+		TerminatingDistinct<Object> consistentWith(QueryScanConsistency scanConsistency);
```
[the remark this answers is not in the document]
daschl: could we just add another consistentWith overload that takes mutation tokens? Then we don't need to break anything and we just carry it forward?

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateQueryIntegrationTests.java
```diff
@@ -148,6 +156,55 @@ void findByMatchingQuery() {
...
+		User specialUser = new User(UUID.randomUUID().toString(), "special", "special");
+
+		couchbaseTemplate.upsertById(User.class).all(Arrays.asList(user1, user2, specialUser));
+
+		Query specialUsers = new Query(QueryCriteria.where("firstname").like("special"));
+		final List<UserJustLastName> foundUsers = couchbaseTemplate.findByQuery(User.class).as(UserJustLastName.class)
+				.consistentWith(QueryScanConsistency.REQUEST_PLUS).matching(specialUsers).all();
+		assertEquals(1, foundUsers.size());
+
+		final List<UserJustLastName> foundUsersReactive = reactiveCouchbaseTemplate.findByQuery(User.class)
```
daschl, line 201: Midway in review I have a basic question.. so here we have:

```
reactiveCouchbaseTemplate.findByQuery(User.class).as(UserJustLastName.class)
```

what is the reason we could not just do 

reactiveCouchbaseTemplate.findByQuery(UserJustLastName.class)

I assume there is good reason, but I probably don't understand it fully

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -109,36 +118,60 @@
...
 		String b = "`" + bucketName + "`";
+		String projectedFields = getProjectedFields(b, resultClass);
 		String entity = "META(" + b + ").id AS " + SELECT_ID + ", META(" + b + ").cas AS " + SELECT_CAS;
 		String count = "COUNT(*) AS " + CountFragment.COUNT_ALIAS;
 		String selectEntity;
-		if (isCount) {
+		if (distinctFields != null) {
+			String distinctFieldsStr = distinctFields.length == 0 ? projectedFields : getDistinctFields(distinctFields);
+			if (isCount) {
+				// only works with one field - MB43475
```
daschl, line 166: can we raise an explicit exception to the user if its more than one field?

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByQueryOperationSupport.java
```diff
@@ -41,26 +42,28 @@ public ExecutableFindByQueryOperationSupport(final CouchbaseTemplate template) {
 
 	@Override
 	public <T> ExecutableFindByQuery<T> findByQuery(final Class<T> domainType) {
-		return new ExecutableFindByQuerySupport<>(template, domainType, ALL_QUERY, QueryScanConsistency.NOT_BOUNDED,
-				"_default._default");
+		return new ExecutableFindByQuerySupport<T>(template, domainType, domainType, ALL_QUERY,
+				QueryScanConsistency.NOT_BOUNDED, null);
```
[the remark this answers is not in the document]
daschl: Yeah, maybe in the worst case we need to come up with a special "null" type of sorts

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByQueryOperationSupport.java
```diff
@@ -115,7 +132,83 @@ public long count() {
 		public boolean exists() {
 			return count() > 0;
 		}
+	}
+
+	/**
+	 * @author Michael Reiche
+	 */
+	static class DistinctOperationSupport<T> implements TerminatingDistinct<T> {
```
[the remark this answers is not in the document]
daschl: Right -- I was wondering something similar .. maybe this is worth discussing with @mp911de before we get this in .. I bet there was a reason and maybe it's specific to mongo or history?

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperation.java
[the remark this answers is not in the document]
mikereiche: That's a good idea.  
In a separate branch, I'll add a withConsistency(QueryScanConsistency) and deprecate the consistentWith(QueryScanConsistency). Then in a future major release, remove it, leaving only the appropriate withConsistency(QueryScanConsistency) and consistentWith(mutation tokens).

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByQueryOperationSupport.java
```diff
@@ -115,7 +132,83 @@ public long count() {
...
+
+		public DistinctOperationSupport(ExecutableFindByQuerySupport<T> delegate, String[] distinctFields) {
+
+			this.delegate = delegate;
+			this.distinctFields = distinctFields;
+		}
+
+		/*
+		 * (non-Javadoc)
+		 * @see org.springframework.data.mongodb.core.ExecutableFindOperation.DistinctWithProjection#as(java.lang.Class)
```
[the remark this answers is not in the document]
mikereiche: oops. I searched all the files in my commit and fixed them.

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateQueryIntegrationTests.java
mikereiche: Good question. Because the documents are stored with _class = "somepackage.User" and they would not be selected by the query.  This is also the reason that there is no as(otherClass) method for ExecutableFindById.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
mikereiche: This was my mistake.  The syntax for a sequence in n1ql is  DISTINCT { prop1, prop2 ... }
I was trying to use SQL syntax of DISTINCT prop1, prop2 ...
I've closed MB43475 and will fix my code.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperation.java
```diff
@@ -46,28 +46,29 @@
...
-	interface FindByIdWithCollection<T> extends TerminatingFindById<T> {
+	interface FindByIdWithCollection<T> {
 
 		/**
 		 * Allows to specify a different collection than the default one configured.
 		 *
 		 * @param collection the collection to use in this scope.
 		 */
-		TerminatingFindById<T> inCollection(String collection);
+		ExecutableFindById<T> inCollection(String collection);
```
mikereiche, line 57: if each with-er returns ExecutableFindById, then there is no need to chain the interfaces.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByIdOperationSupport.java
```diff
@@ -63,13 +62,13 @@ public T one(final String id) {
...
-		public TerminatingFindById<T> inCollection(final String collection) {
+		public ExecutableFindById<T> inCollection(final String collection) {
 			Assert.hasText(collection, "Collection must not be null nor empty.");
 			return new ExecutableFindByIdSupport<>(template, domainType, collection, fields);
 		}
 
 		@Override
-		public FindByIdWithCollection<T> project(String... fields) {
+		public ExecutableFindById<T> project(String... fields) {
 			Assert.notEmpty(fields, "Fields must not be null nor empty.");
```
mikereiche, line 72: I find it strange that FindById does projection by project(fields) instead of as(returnType) like everything else.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByQueryOperation.java
```diff
@@ -107,40 +108,90 @@
...
+		/**
+		 * Set the filter {@link QueryCriteriaDefinition criteria} to be used.
+		 *
+		 * @param criteria must not be {@literal null}.
+		 * @return new instance of {@link ExecutableFindByQuery}.
+		 * @throws IllegalArgumentException if criteria is {@literal null}.
+		 */
+		default ExecutableFindByQuery<T> matching(QueryCriteriaDefinition criteria) {
+			return matching(Query.query(criteria));
+		}
```
mikereiche, line 130: mongo provides a matching(Criteria) method.

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByIdOperationSupport.java
```diff
@@ -62,6 +62,8 @@
 		public Mono<T> one(final String id) {
 			return Mono.just(id).flatMap(docId -> {
 				GetOptions options = getOptions().transcoder(RawJsonTranscoder.INSTANCE);
+				// Instead of project(fields), I think this should be implemented the same as
+				// FindByQuery as(resultType), and the list of fields should be the properties of resultType
```
mikereiche, line 66: Here is the run-time for FindById.project() that I questioned above.

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
```diff
@@ -102,49 +124,61 @@ public ReactiveFindByQueryOperationSupport(final ReactiveCouchbaseTemplate templ
...
+					String id = "";
+					long cas = 0;
+					if (distinctFields == null) {
+						id = row.getString(TemplateUtils.SELECT_ID);
+						cas = row.getLong(TemplateUtils.SELECT_CAS);
+						row.removeKey(TemplateUtils.SELECT_ID);
+						row.removeKey(TemplateUtils.SELECT_CAS);
+					}
+					return template.support().decodeEntity(id, row.toString(), cas, returnType);
+				});
```
mikereiche, line 149: If a collection has been specified, execute on the scope, otherwise execute on the cluster.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperation.java
```diff
@@ -114,17 +114,30 @@
...
+		/**
+		 * Allows to override the default scan consistency.
+		 *
+		 * @param scanConsistency the custom scan consistency to use for this analytics query.
+		 */
+		FindByAnalyticsWithQuery<T> withConsistency(AnalyticsScanConsistency scanConsistency);
+
+	}
+
+	interface ExecutableFindByAnalytics<T> extends FindByAnalyticsConsistentWith<T>, FindByAnalyticsWithConsistency<T> {}
```
mikereiche, line 141: Deprecate consistentWith(), add withConsistency(). This is done for all interfaces that have consistentWith()

daschl (approved):
looks good to me, the only thing I'm unsure on are the interface extension changes -- please triple check that those are not breaking any users out there...

PR #1051, opened 2021-01-15
DATACOUCH-1041 - Add withCas() method to RemoveById.

mikereiche:
Add withCas() method to RemoveById as the cas in the entity cannot be used as the entity is not pass in the API.  The other option was to add methods like one(T entity) and all(Collection<T>), but those class with one(String id) and all(Collection<String> ids).

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperation.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperation.java
@@ -50,6 +50,11 @@ public interface ExecutableRemoveByIdOperation {
 
 	}
 
-	interface ExecutableRemoveById extends RemoveByIdWithDurability {}
+	interface RemoveByIdWithCas extends RemoveByIdWithDurability {
+
+		RemoveByIdWithDurability withCas(Long cas);
+	}
+
+	interface ExecutableRemoveById extends RemoveByIdWithCas {}
 
 }
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperationSupport.java
@@ -35,7 +35,8 @@ public class ExecutableRemoveByIdOperationSupport implements ExecutableRemoveByI
 
 	@Override
 	public ExecutableRemoveById removeById() {
-		return new ExecutableRemoveByIdSupport(template, null, PersistTo.NONE, ReplicateTo.NONE, DurabilityLevel.NONE);
+		return new ExecutableRemoveByIdSupport(template, null, PersistTo.NONE, ReplicateTo.NONE, DurabilityLevel.NONE,
+				null);
 	}
 
 	static class ExecutableRemoveByIdSupport implements ExecutableRemoveById {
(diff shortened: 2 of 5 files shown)


src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperation.java
```diff
@@ -50,6 +50,11 @@
 
 	}
 
-	interface ExecutableRemoveById extends RemoveByIdWithDurability {}
+	interface RemoveByIdWithCas extends RemoveByIdWithDurability {
+
+		RemoveByIdWithDurability withCas(Long cas);
```
daschl, line 55: why is this a boxed long and not an unboxed `long`?

src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperationSupport.java
```diff
@@ -45,17 +46,19 @@ public ExecutableRemoveById removeById() {
 		private final PersistTo persistTo;
 		private final ReplicateTo replicateTo;
 		private final DurabilityLevel durabilityLevel;
+		private final Long cas;
```
daschl, line 49: same here, could this be a primitive long?

src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByIdOperation.java
```diff
@@ -52,6 +53,12 @@
 
 	}
 
-	interface ReactiveRemoveById extends RemoveByIdWithDurability {}
+	interface RemoveByIdWithCas extends RemoveByIdWithDurability {
+
+		RemoveByIdWithDurability withCas(Long cas);
```
daschl, line 58: see above w.r.t boxing

src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByIdOperationSupport.java
```diff
@@ -81,28 +83,36 @@ private RemoveOptions buildRemoveOptions() {
 			} else if (durabilityLevel != DurabilityLevel.NONE) {
 				options.durability(durabilityLevel);
 			}
+			if (cas != null) {
```
daschl, line 86: I see the boxing and null, but it could just check if != 0 ?

daschl (changes requested):
looks good to me overall, but I think we can get rid of the boxing for less gc

src/main/java/org/springframework/data/couchbase/core/ExecutableRemoveByIdOperation.java
mikereiche: So that null can have different semantics than zero.

src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByIdOperationSupport.java
mikereiche: This comes from a comment made by Mark Paluch on another change-set regarding Expiry(I think). It's not quite the same for CAS, but I wanted to keep the same pattern/semantics - that null means "never been set/not specified", while an Object (including an object with value 0) means that it has been set.

https://github.com/spring-projects/spring-data-couchbase/pull/292#pullrequestreview-549850646

In other words - something like this -  That if it was set to non-Zero by an annotation, then there was no way to override (unset) it with the fluent call because it was already non-Duration.zero and the fluent call had to set it to Duration.zero, but Duration.zero (from the fluent call or otherwise) is treated as "never been set", so the value from the annotation would not be overridden. So null would have the semantics of "never been set", and Duration.zero has the semantics of "been set to zero".

PR #1082, opened 2021-02-17
DATACOUCH-1066 - Make QueryCriteria key a N1qlExpression.

### LINKED ISSUE #1066: QueryCriteria.where() adds backticks to ALL keys, breaks keys that use META()
For example:
```java
import static org.springframework.data.couchbase.core.query.N1QLExpression.*;
...
QueryCriteria criteria = QueryCriteria.where(path(meta(escapedBucket(getCouchbaseOperations().getBucketName())), "cas").toString()).eq(version);
```

produces a query like so:

```
(`META(`YS_Apps`).cas` = 1611287177404088320)
```

mikereiche:
This is a merge of mmonti's changeset into master. The changeset was made
on top of 4.1.x instead of master so it has a little catching up to do.

Closes #1066

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/query/StringN1qlQueryCreator.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/StringN1qlQueryCreator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 the original author or authors
+ * Copyright 2021 the original author or authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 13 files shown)


src/main/java/org/springframework/data/couchbase/repository/query/StringN1qlQueryCreator.java
```diff
@@ -52,6 +55,7 @@
 	private final StringBasedN1qlQueryParser queryParser;
 	private final QueryMethod queryMethod;
 	private final CouchbaseConverter couchbaseConverter;
+	private static final SpelExpressionParser SPEL_PARSER = new SpelExpressionParser();
 	private final N1QLExpression parsedExpression;
```
mikereiche: Never used. Removing.

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateQueryIntegrationTests.java
```diff
@@ -124,7 +126,7 @@ void findByMatchingQuery() {
 
 		couchbaseTemplate.upsertById(User.class).all(Arrays.asList(user1, user2, specialUser));
 
-		Query specialUsers = new Query(QueryCriteria.where("firstname").like("special"));
+		Query specialUsers = new Query(QueryCriteria.where(i("firstname")).like("special"));
```
aaronjwhiteside, line 129: It might already be in the PR, but I wasn't paying close enough attention

But can the value methods take `N1QLExpression`'s to? I can imagine that somewhere someone might want to use a built in N1QL method to compute the value to compare against..

PR #1089, opened 2021-02-18
Manually finish the revert of datacouch_550.

### LINKED ISSUE #858: AutoIndexing is not Working properly [DATACOUCH-550]
**[deniswsrosa](https://jira.spring.io/secure/ViewProfile.jspa?name=deniswsrosa)** opened **[DATACOUCH-550](https://jira.spring.io/browse/DATACOUCH-550?redirect=false)** and commented

I have enabled autoindexing according to the docs:

```java
@Configuration
public class Config extends AbstractCouchbaseConfiguration {


 @Override
 public String getConnectionString() {
 return "couchbase://127.0.0.1";
 }

 @Override
 public String getUserName() {
 return "Administrator";
 }

 @Override
 public String getPassword() {
 return "password";
 }

 @Override
 public String getBucketName() {
 return "default";
 }


 /**
 * OPTIONAL: In this example we are using annotations to
 */

 @Override
 protected boolean autoIndexCreation() {
 return true;
 }
}
```

 

And here is my entity:

 

```java
@Document
@Data
@ToString
@CompositeQueryIndex(fields = {"id", "name", "position"})
public class User {

 @Id
 private String id;
 @QueryIndexed
 private String name;
 @QueryIndexed
 private String position;
 private String description;
 private double accountBalance;
}
```

 

I have tried many variations of the mapping above, and after the application starts no index is created on the target bucket.

 

 

 

 

 

---

**Affects:** 4.0 GA (Neumann)

**Referenced from:** pull request https://github.com/spring-projects/spring-data-couchbase/pull/295

1 votes, 5 watchers

mikereiche:
The previous "Revert" only reverted CouchbaseTemplate resulting in compilation errors in master.

Closes #858

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -37,7 +37,6 @@ import org.springframework.data.couchbase.core.convert.CouchbaseCustomConversion
 import org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter;
 import org.springframework.data.couchbase.core.convert.translation.JacksonTranslationService;
 import org.springframework.data.couchbase.core.convert.translation.TranslationService;
-import org.springframework.data.couchbase.core.index.CouchbasePersistentEntityIndexCreator;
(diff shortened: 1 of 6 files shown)


daschl:
merged into master

PR #1102, opened 2021-03-17
DATACOUCH-955 - Add support for reactive auditing and ReactiveEntityCallbacks.

### LINKED ISSUE #955: Add @EnableReactiveCouchbaseAuditing  [DATACOUCH-644]
**[hantsy](https://jira.spring.io/secure/ViewProfile.jspa?name=hantsy1)** opened **[DATACOUCH-644](https://jira.spring.io/browse/DATACOUCH-644?redirect=false)** and commented

In this new Spring Data commons, there is a new ReactiveAuitorAware added for reactive stack.

The R2dbc, Mongo, Neo4j, etc had added a variant to enable this support.

Suggest use  `@EnableReactiveCouchbaseAuditing` to enable it

---

**Affects:** 4.1 GA (2020.0.0)

**Reference URL:** https://github.com/hantsy/spring-reactive-sample/tree/master/boot-data-couchbase

jorgerod:
Closes #955.
Original pull request: #1111.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors
+ * Copyright 2012-2021 the original author or authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,7 +53,7 @@ public class ExecutableFindByAnalyticsOperationSupport implements ExecutableFind
 			this.domainType = domainType;
 			this.query = query;
 			this.reactiveSupport = new ReactiveFindByAnalyticsSupport<>(template.reactive(), domainType, query,
-					scanConsistency);
+					scanConsistency, new NonReactiveSupportWrapper(template.support()));
 			this.scanConsistency = scanConsistency;
 		}
 
--- a/src/main/java/org/springframework/data/couchbase/core/NonReactiveSupportWrapper.java
+++ b/src/main/java/org/springframework/data/couchbase/core/NonReactiveSupportWrapper.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.couchbase.core;
+
+import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
+
+import reactor.core.publisher.Mono;
+
+/**
+ * Wrapper of {@link TemplateSupport} methods to adapt them to {@link ReactiveTemplateSupport}.
+ *
+ * @author Carlos Espinaco
+ * @since 4.2
+ */
+public class NonReactiveSupportWrapper implements ReactiveTemplateSupport {
+
+	private final TemplateSupport support;
+
+	public NonReactiveSupportWrapper(TemplateSupport support) {
+		this.support = support;
+	}
+
+	@Override
+	public Mono<CouchbaseDocument> encodeEntity(Object entityToEncode) {
+		return Mono.fromSupplier(() -> support.encodeEntity(entityToEncode));
+	}
+
+	@Override
+	public <T> Mono<T> decodeEntity(String id, String source, long cas, Class<T> entityClass) {
+		return Mono.fromSupplier(() -> support.decodeEntity(id, source, cas, entityClass));
+	}
+
+	@Override
+	public Mono<Object> applyUpdatedCas(Object entity, long cas) {
+		return Mono.fromSupplier(() -> support.applyUpdatedCas(entity, cas));
+	}
+
+	@Override
+	public Mono<Object> applyUpdatedId(Object entity, Object id) {
+		return Mono.fromSupplier(() -> support.applyUpdatedId(entity, id));
+	}
+
+	@Override
+	public Long getCas(Object entity) {
+		return support.getCas(entity);
+	}
+
+	@Override
+	public String getJavaNameForEntity(Class<?> clazz) {
+		return support.getJavaNameForEntity(clazz);
+	}
+}
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
@@ -1,0 +1,222 @@
+/*
+ * Copyright 2012-2021 the original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.data.couchbase.core;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.data.couchbase.core.convert.CouchbaseConverter;
+import org.springframework.data.couchbase.core.convert.translation.TranslationService;
+import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
+import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
+import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
+import org.springframework.data.couchbase.core.mapping.event.BeforeConvertEvent;
+import org.springframework.data.couchbase.core.mapping.event.BeforeSaveEvent;
+import org.springframework.data.couchbase.core.mapping.event.CouchbaseMappingEvent;
+import org.springframework.data.couchbase.core.mapping.event.ReactiveAfterConvertCallback;
+import org.springframework.data.couchbase.core.mapping.event.ReactiveBeforeConvertCallback;
+import org.springframework.data.couchbase.repository.support.MappingCouchbaseEntityInformation;
+import org.springframework.data.mapping.PersistentPropertyAccessor;
+import org.springframework.data.mapping.callback.EntityCallbacks;
+import org.springframework.data.mapping.callback.ReactiveEntityCallbacks;
+import org.springframework.data.mapping.context.MappingContext;
+import org.springframework.data.mapping.model.ConvertingPropertyAccessor;
+import org.springframework.util.Assert;
+import reactor.core.publisher.Mono;
+
+/**
+ * Internal encode/decode support for {@link ReactiveCouchbaseTemplate}.
+ *
+ * @author Carlos Espinaco
+ * @since 4.2
+ */
+class ReactiveCouchbaseTemplateSupport implements ApplicationContextAware, ReactiveTemplateSupport {
+
+	private static final Logger LOG = LoggerFactory.getLogger(ReactiveCouchbaseTemplateSupport.class);
+
+	private final CouchbaseConverter converter;
+	private final MappingContext<? extends CouchbasePersistentEntity<?>, CouchbasePersistentProperty> mappingContext;
+	private final TranslationService translationService;
+	private ReactiveEntityCallbacks reactiveEntityCallbacks;
+	private ApplicationContext applicationContext;
+
+	public ReactiveCouchbaseTemplateSupport(final CouchbaseConverter converter,
+			final TranslationService translationService) {
+		this.converter = converter;
+		this.mappingContext = converter.getMappingContext();
+		this.translationService = translationService;
+	}
+
+	@Override
+	public Mono<CouchbaseDocument> encodeEntity(final Object entityToEncode) {
+		return Mono.just(entityToEncode)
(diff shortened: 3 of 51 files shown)


src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
```diff
@@ -53,7 +53,7 @@ public ExecutableFindByAnalyticsOperationSupport(final CouchbaseTemplate templat
 			this.domainType = domainType;
 			this.query = query;
 			this.reactiveSupport = new ReactiveFindByAnalyticsSupport<>(template.reactive(), domainType, query,
-					scanConsistency);
+					scanConsistency, new NonReactiveSupportWrapper(template.support()));
```
mikereiche, line 56: Isn't this NonReactiveSupportWrapper object the same as template.reactive().support() ? 
Then this extra argument  to the Reactive*Support constructors is not necessary.
And the ReactiveTemplateSupport support object in the Reactive*Support objects is not necessary - they can just use template.support().

src/main/java/org/springframework/data/couchbase/core/NonReactiveSupportWrapper.java
```diff
@@ -0,0 +1,65 @@
+/*
```
mikereiche, line 1: This class isn't necessary.  template.support() of the Reactive*OperationSupport classes will do the same thing.

src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
```diff
@@ -54,28 +54,28 @@ public ReactiveInsertByIdOperationSupport(final ReactiveCouchbaseTemplate templa
...
+			return (Mono<T>) Mono.just(object).flatMap(support::encodeEntity).flatMap(converted -> {
 				return template.getCollection(collection).reactive()
-						.insert(converted.getId(), converted.export(), buildInsertOptions(converted)).map(result -> {
-							Object updatedObject = template.support().applyUpdatedId(o, converted.getId());
-							return (T) template.support().applyUpdatedCas(updatedObject, result.cas());
-						});
+						.insert(converted.getId(), converted.export(), buildInsertOptions(converted)).flatMap(result ->
+							support.applyUpdatedId(object, converted.getId())
+								.map(updatedObject -> support.applyUpdatedCas(updatedObject, result.cas())));
 			}).onErrorMap(throwable -> {
```
mikereiche, line 79: I had test failures here with.  src/test/resources/integration.properties, cluster.type=unmanaged 
mvn integration-test
I think this code needs to be as follows (similar to ReactiveUpsertByIdOperationSupport.one()

```
public Mono<T> one(T object) {
	return (Mono<T>)Mono.just(object).flatMap(template.support()::encodeEntity).flatMap(converted ->
			template.getCollection(collection).reactive()
				.insert(converted.getId(), converted.export(), buildInsertOptions(converted)).flatMap(result ->
					template.support().applyUpdatedId(object, converted.getId())
							.flatMap( updatedObject -> template.support().applyUpdatedCas(updatedObject, result.cas())))
	).onErrorMap(throwable -> {
		if (throwable instanceof RuntimeException) {
			return template.potentiallyConvertRuntimeException((RuntimeException) throwable);
		} else {
			return throwable;
		}
	});
}
```

mikereiche (changes requested):
It appears that the net result is that instead of CouchbaseTemplateSupport methods being called, now the corresponding ReactiveCouchbaseTemplateSupport methods are being called, resulting in the auditing configured as ReactiveAuditing being called instead of the the (nonReactive)auditing being called.  And CouchbaseTemplateSupport is no longer used.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
jorgerod: Hi @mikereiche 

In my opinion and following the implementations of other spring-data projects such as spring-data-mongo, I believe that in  blocked stack we should launch `EntityCallback` and from the reactive stack we should launch `ReactiveEntityCallback`.
To have that differentiation, there needs to be one path for blocking (`CouchbaseTemplateSupport`) and another path for reactive (`ReactiveCouchbaseTemplateSupport`).

Also, I think that, additionally, the blocking audit event should be linked to `AuditingEntityCallback` instead of `AuditingEventListener` (this is not in the PR but you could add it).

How do you see it?

mikereiche: Got it. Thanks for the explanation.

src/main/java/org/springframework/data/couchbase/core/mapping/event/ReactiveAuditingEntityCallback.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * Reactive {@link EntityCallback} to populate auditing related fields on an entity about to be
+ * saved. Based on <a href="https://github.com/spring-projects/spring-data-mongodb/blob/3.1.6/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/event/ReactiveAfterConvertCallback.java">ReactiveAfterConvertCallback</a>
+ *
+ * @author Jorge Rodríguez Martín
+ * @since 4.2
+ */
+public class ReactiveAuditingEntityCallback implements ReactiveBeforeConvertCallback<Object>,
```
mikereiche, line 34: Make this also implement ReactiveAfterConvertCallback<Object>.  Add onAfterConvert(Object, CouchbaseDocument, String)

src/main/java/org/springframework/data/couchbase/core/mapping/event/ReactiveAfterConvertCallback.java
```diff
@@ -0,0 +1,40 @@
...
+public interface ReactiveAfterConvertCallback<T> extends EntityCallback<T> {
+
+	/**
+	 * Entity callback method invoked after a domain object is converted to be persisted. Can return
+	 * either the same of a modified instance of the domain object.
+	 * @param entity the domain object to save.
+	 * @param collection name of the collection.
+	 * @return a {@link Publisher} emitting the domain object to be persisted.
+	 */
+	Publisher<T> onAfterConvert(T entity, Document document, String collection);
```
mikereiche, line 39: I think this should be CouchbaseDocument (org.springframework.data.couchbase.core.mapping.Document is an annotation)

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
jorgerod: This point
> Also, I think that, additionally, the blocking audit event should be linked to AuditingEntityCallback instead of AuditingEventListener (this is not in the PR but you could add it).

Do you want me to add it to this PR?

mikereiche (commented):
There are four fixes in comments, then I will approve and merge it.
It can be manually tested by adding the @EnableReactiveCouchbaseAuditing annotation to ReactiveCouchbaseRepositoryKeyValueIntegrationTests.Config
And the @EnableCouchbaseAuditing annotation to CouchbaseRepositoryKeyValueIntegrationTests.Config
then changing src/test/resources/integration.properties
cluster.type=unmanaged

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
mikereiche: Yes please.

src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
jorgerod: Hi @mikereiche 

This test fails because if you enable auditing, field annoted with @CreatedBy, @CreatedDate, @LastModifiedBy and @LastModifiedDate are injected.

IMHO, the test is not entirely correct as the assertion should be between the User obtained from `userRepository.save(user).block()` and the User obtained from userRepository.findById(user.getId()).blockOptional().

src/main/java/org/springframework/data/couchbase/core/mapping/event/ReactiveAuditingEntityCallback.java
jorgerod: I think it is not correct to add `ReactiveAfterConvertCallback`. Audit event should only be associated to a single event (ReactiveBeforeConvertCallback).

If we also add the `ReactiveAfterConvertCallback`, in each conversion the information of the fields annotated with @CreatedBy, @CreatedDate, @LastModifiedBy and @LastModifiedDate will be modified 2 times and the returned object will have different values for the date type fields than the saved CouchbaseDocument.

src/main/java/org/springframework/data/couchbase/core/ExecutableFindByAnalyticsOperationSupport.java
jorgerod: Done

src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
```diff
@@ -0,0 +1,222 @@
...
+			}
+		} else {
+			LOG.info("maybeCallBeforeConvert called, but ReactiveCouchbaseTemplate not initialized with applicationContext");
+		}
+		return Mono.just(object);
+	}
+
+	protected <T> Mono<T> maybeCallAfterConvert(T object, CouchbaseDocument document, String collection) {
+		if (null != reactiveEntityCallbacks) {
+			return reactiveEntityCallbacks.callback(AfterConvertCallback.class, object, document, collection);
```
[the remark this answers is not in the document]
jorgerod: Done

src/main/java/org/springframework/data/couchbase/core/mapping/event/ReactiveAfterConvertCallback.java
jorgerod: Done

src/main/java/org/springframework/data/couchbase/core/ReactiveInsertByIdOperationSupport.java
mikereiche: Please set src/test/resources/integration.properties, cluster.type=unmanaged
and run mvn integration-test.  It gives ClassCastExceptions.
Yes, it should be comparing the entities from save and find. 

[ERROR] Errors: 
[ERROR]   CouchbaseTemplateKeyValueIntegrationTests.existsById:251 ClassCast reactor.cor...
[ERROR]   CouchbaseTemplateKeyValueIntegrationTests.insertById:231 ClassCast reactor.cor...
[ERROR]   CouchbaseTemplateKeyValueIntegrationTests.insertByIdwithDurability:240 ClassCast
[ERROR]   CouchbaseTemplateKeyValueIntegrationTests.saveAndFindImmutableById:266 ClassCast
[ERROR]   CouchbaseTemplateKeyValueIntegrationTests.withDurability:115 ClassCast reactor...
[ERROR]   CouchbaseTemplateKeyValueIntegrationTests.withExpiryAndExpiryAnnotation:157 ClassCast

jorgerod: Fixed

mikereiche:
Couple little things - they changed the format for Title/Description.   The no longer want the DATACOUCH-955 in the title, it should just be "Add support for reactive auditing and ReactiveEntityCallbacks"
And in the description where you have "Fixes #955" it should be as below (Closes, Original pull request:)

Closes #955.
Original pull request: #1102.

I'll have one of my colleagues review it as well before merging.

mikereiche:
@jorgerod -

Can you add a test in ReactiveCouchbaseRepositoryQueryIntegrationTests?  And an analogous test in CouchbaseRepositoryIntegrationTests to check for regression.

Add createdBy to Airport

	@CreatedBy String createdBy;
	public String getCreatedBy(){
		return createdBy;
	}

add

	@Test
	void findBySimplePropertyAudited() {
		Airport vie = null;
		try {
			vie = new Airport("airports::vie", "vie", "low2");
			Airport saved = airportRepository.save(vie).block();
			List<Airport> airports1 = airportRepository.findAllByIata("vie").collectList().block();
			assertEquals(saved, airports1.get(0));
			assertEquals(saved.getCreatedBy(), "auditor"); // NaiveAuditorAware will provide this
		} finally {
			airportRepository.delete(vie).block();
		}
	}

Updated the Config class in ReactiveCouchbaseRepositoryQueryIntegrationTests.  NaiveAuditorAware and AuditingDateTimeProvider already exist.

	@Configuration
	@EnableReactiveCouchbaseRepositories("org.springframework.data.couchbase")
	@EnableCouchbaseAuditing(auditorAwareRef = "auditorAwareRef", dateTimeProviderRef = "dateTimeProviderRef")
	static class Config extends AbstractCouchbaseConfiguration {

		@Override
		public String getConnectionString() {
			return connectionString();
		}

		@Override
		public String getUserName() {
			return config().adminUsername();
		}

		@Override
		public String getPassword() {
			return config().adminPassword();
		}

		@Override
		public String getBucketName() {
			return bucketName();
		}

		@Bean(name = "auditorAwareRef")
		public NaiveAuditorAware testAuditorAware() {
			return new NaiveAuditorAware();
		}

		@Bean(name = "dateTimeProviderRef")
		public DateTimeProvider testDateTimeProvider() {
			return new AuditingDateTimeProvider();
		}

	}

jorgerod:
@mikereiche I have added the tests you asked for

src/main/java/org/springframework/data/couchbase/repository/auditing/CouchbaseAuditingRegistrar.java
```diff
@@ -84,7 +86,7 @@ protected void registerAuditListenerBeanDefinition(BeanDefinition auditingHandle
 				.addConstructorArgValue(ParsingUtils.getObjectFactoryBeanDefinition(getAuditingHandlerBeanName(), registry));
 
 		registerInfrastructureBeanWithId(listenerBeanDefinitionBuilder.getBeanDefinition(),
-				AuditingEventListener.class.getName(), registry);
+				AuditingEntityCallback.class.getName(), registry);
```
mikereiche, line 89: I think this just changes the name used in registering?  On line 84, the rootBeanDefinitions is still AuditingEventListener.class - and after this change, the AuditEventListener is still the listener class is being used.   And it already had an onApplicationEvent() that calls markedAudit().   The changed line above can even be changed to as below and it still works.  So I don't think this particular change is needed - and I don't think the AuditingEventCallback class is needed.

/*AuditingEntityCallback.class.getName()*/ "My_Friend_Dave", registry

mikereiche: ok - I just looked at the ReactiveCouchbaseAuditingRegistrar - and it appears that the intention is to use the new AuditEntityCallback.class - both when creating the listenerBeanDefinitionBuilder and when registering.   And to no longer use the AuditEventListener.

PR #1132, opened 2021-04-23
Expose ObjectMapper used by JsonSerializer

### LINKED ISSUE #1130: add couchbaseJsonValueModule bean to AbstractCouchbaseConfiguration
https://forums.couchbase.com/t/setting-json-deserialisation-features-in-java-sdk-3/30215/6

We Could change contentAsObject() to always use JacksonTransformers.bytesToJsonObject() so it doesn't depend on the user registering the correct module. What do you think?
Should also add another preflight check and fail if JsonValueModule isn't registered.

mikereiche:
Expose ObjectMapper used by JsonSerializer

Closes #1130.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -16,12 +16,13 @@
 
 package org.springframework.data.couchbase.config;
 
-import static com.couchbase.client.java.ClusterOptions.*;
+import static com.couchbase.client.java.ClusterOptions.clusterOptions;
 
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
@@ -48,11 +49,19 @@ import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
 import com.couchbase.client.core.deps.com.fasterxml.jackson.databind.DeserializationFeature;
+import com.couchbase.client.core.deps.com.fasterxml.jackson.databind.Module;
+import com.couchbase.client.core.encryption.CryptoManager;
 import com.couchbase.client.core.env.Authenticator;
 import com.couchbase.client.core.env.PasswordAuthenticator;
+import com.couchbase.client.core.error.CouchbaseException;
 import com.couchbase.client.java.Cluster;
+import com.couchbase.client.java.codec.JacksonJsonSerializer;
+import com.couchbase.client.java.encryption.databind.jackson.EncryptionModule;
 import com.couchbase.client.java.env.ClusterEnvironment;
 import com.couchbase.client.java.json.JacksonTransformers;
+import com.couchbase.client.java.json.JsonValueModule;
+import com.couchbase.client.java.json.RepackagedJsonValueModule;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
  * Base class for Spring Data Couchbase configuration using JavaConfig.
@@ -65,12 +74,13 @@ import com.couchbase.client.java.json.JacksonTransformers;
 @Configuration
 public abstract class AbstractCouchbaseConfiguration {
 
+	@Autowired ObjectMapper couchbaseObjectMapper;
+
 	/**
 	 * The connection string which allows the SDK to connect to the cluster.
 	 * <p>
-	 * Note that the connection string can take many forms, in its simplest it is just a single hostname
-	 * like "127.0.0.1". Please refer to the couchbase Java SDK documentation for all the different
-	 * possibilities and options.
+	 * Note that the connection string can take many forms, in its simplest it is just a single hostname like "127.0.0.1".
+	 * Please refer to the couchbase Java SDK documentation for all the different possibilities and options.
 	 */
 	public abstract String getConnectionString();
 
@@ -130,6 +140,10 @@ public abstract class AbstractCouchbaseConfiguration {
 	@Bean(destroyMethod = "shutdown")
 	public ClusterEnvironment couchbaseClusterEnvironment() {
 		ClusterEnvironment.Builder builder = ClusterEnvironment.builder();
+		if (!nonShadowedJacksonPresent()) {
+			throw new CouchbaseException("non-shadowed Jackson not present");
+		}
+		builder.jsonSerializer(JacksonJsonSerializer.create(couchbaseObjectMapper));
 		configureEnvironment(builder);
 		return builder.build();
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -130,6 +140,10 @@ public Cluster couchbaseCluster(ClusterEnvironment couchbaseClusterEnvironment)
 	@Bean(destroyMethod = "shutdown")
 	public ClusterEnvironment couchbaseClusterEnvironment() {
 		ClusterEnvironment.Builder builder = ClusterEnvironment.builder();
+		if (!nonShadowedJacksonPresent()) {
+			throw new CouchbaseException("non-shadowed Jackson not present");
+		}
+		builder.jsonSerializer(JacksonJsonSerializer.create(couchbaseObjectMapper));
```
mikereiche, line 146: use the couchbaseObjectMapper from the new Bean.
Check for nonShadowedJacksonPresent() although it should never happen.

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -273,6 +287,31 @@ public CouchbaseMappingContext couchbaseMappingContext(CustomConversions customC
...
+		ObjectMapper mapper = new ObjectMapper();
+		mapper.configure(com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+		mapper.registerModule(new JsonValueModule());
+		CryptoManager cryptoManager = null;
+		if (cryptoManager != null) {
+			mapper.registerModule(new EncryptionModule(cryptoManager));
+		}
+		return mapper;
+	}
+
```
mikereiche, line 314: the couchbaseObjectMapper that gets used and exposed. Created the same as in ClusterEnvironment.Builder.

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -184,6 +184,7 @@ void findByTypeAlias() {
 			vie = new Airport("airports::vie", "vie", "loww");
 			vie = airportRepository.save(vie);
 			List<Airport> airports = couchbaseTemplate.findByQuery(Airport.class)
+					.withConsistency(QueryScanConsistency.REQUEST_PLUS)
```
mikereiche, line 187: fixed a test that was sporadically failing.

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -198,7 +199,7 @@ void findByEnum() {
 			vie = new Airport("airports::vie", "vie", "loww");
 			vie = airportRepository.save(vie);
 			Airport airport2 = airportRepository.findByIata(Iata.vie);
-			assertNotNull(airport2, "should have found "+vie);
+			assertNotNull(airport2, "should have found " + vie);
```
mikereiche, line 202: formatting.  Nothing to see.

src/test/java/org/springframework/data/couchbase/repository/ReactiveCouchbaseRepositoryKeyValueIntegrationTests.java
```diff
@@ -64,13 +66,13 @@ void saveAndFindById() {
...
-	void findBySimplePropertyAudited() {
+	void findByIdAudited() {
 		Airport vie = null;
 		try {
 			vie = new Airport("airports::vie", "vie", "low2");
 			Airport saved = airportRepository.save(vie).block();
-			List<Airport> airports1 = airportRepository.findAllByIata("vie").collectList().block();
-			assertEquals(saved, airports1.get(0));
+			Airport airport1 = airportRepository.findById(saved.getId()).block();
+			assertEquals(airport1, saved);
```
mikereiche, line 75: This was using a query which requires an index, but if you run this test by itself, there won't be an index.  Switch to use kv.

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -130,6 +140,10 @@ public Cluster couchbaseCluster(ClusterEnvironment couchbaseClusterEnvironment)
 	@Bean(destroyMethod = "shutdown")
 	public ClusterEnvironment couchbaseClusterEnvironment() {
 		ClusterEnvironment.Builder builder = ClusterEnvironment.builder();
+		if (!nonShadowedJacksonPresent()) {
+			throw new CouchbaseException("non-shadowed Jackson not present");
```
daschl, line 144: is there a reason we need to blow up if it is not present? (or just add the custom serializer if present?) -  could log a warning instead if unlikely?

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -273,6 +287,31 @@ public CouchbaseMappingContext couchbaseMappingContext(CustomConversions customC
...
+	/**
+	 * Creates a {@link Module} ObjectMapper mapper = new ObjectMapper();
+	 * mapper.registerModule(couchbaseJsonValueModule()); <----- ClusterEnvironment env = ClusterEnvironment.builder()
+	 * .jsonSerializer(new JacksonJsonSerializer(mapper)) .build();
+	 *
+	 * @throws Exception on Bean construction failure.
+	 * @return
+	 */
+	@Bean
+	public com.couchbase.client.core.deps.com.fasterxml.jackson.databind.Module couchbaseJsonValueModule() {
```
daschl, line 299: does this have to be its own bean?

daschl (commented):
Some clarifications/input

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
mikereiche: Good question. In the SDK, when the non-shadowed jackson classes are not present, a DefaultJsonSerializer is created.  DefaultJsonSerializer has no means to use an already-created mapper, or expose the mapper that it creates.  I suppose this would only be an issue when the user wants to us couchbaseObjectMapper() and expects it to be the same one used by the serializer.  However, David said the not having the non-shadowed classes in Spring is impossible.  The alternative would be to issue a warning saying that the serializer was not using couchbaseObjectMapper() and continue.

   return nonShadowedJacksonPresent()
        ? JacksonJsonSerializer.create(cryptoManager)
        : DefaultJsonSerializer.create(cryptoManager);

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
mikereiche: This doesn't need to exist at all. Removed.

PR #1135, opened 2021-05-03
Add override of toN1qlRemoveString() to StringQuery.

### LINKED ISSUE #1131: Delete N1QL query is deleting all documents instead of deleting only documents that matches the criteria
I have a simple requirement to delete all documents whose date is less than specified in the query. but whatever option I try, it’s deleting all documents from db. Code is written in Java Spring Boot project.  
versions:  
Sprinboot-2.4.4.   
spring data couchbase-4.1.6  
couchbase server: 6  

Sample codes I tried: I tried passing date argument as String, LocalDateTime, etc but none of it worked.
```
@Query("#{#n1ql.delete} WHERE #{#n1ql.filter} AND dateParam < MILLIS($1)")
void selectDateTime(LocalDateTime/String dateParam);

```
```
@Query("#{#n1ql.delete} WHERE #{#n1ql.filter} AND dateParam < $1")
void selectDateTime(String dateParam);
```

```
@Query("#{#n1ql.delete} WHERE #{#n1ql.filter} AND dateParam < MILLIS($1) #{#n1ql.returning}")
List selectDateTime(String dateParam);
```

dateParam passed as String was: “2021-04-07T06:47:32Z”
I also tried passing LocalDateTime- LocalDateTime.of(2021, 4, 7, 06, 47, 32);

To confirm there is nothing wrong in the criteria or data type or values I'm passing to above queries, I tried select query with same where clause and it's working perfectly fine - correct number of documents matching criteria is returned.

```
@Query("#{#n1ql.selectEntity} WHERE #{#n1ql.filter} AND syncTimeStamp < MILLIS($1)")
void selectDateTime(LocalDateTime/String date);
```

It’s surprising it’s working for select query but not for delete query.

If I run query from couchbase UI, it works fine - only documents that matches condition gets deleted.
`delete from DB_NAME where dateParam < 1618188052352`

Any help would be appreciated.

mikereiche:
Closes #1131.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/query/Query.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/Query.java
@@ -323,7 +323,10 @@ public class Query {
 	 * @return QueryOptions
 	 */
 	public QueryOptions buildQueryOptions(QueryScanConsistency scanConsistency) {
-		final QueryOptions options = QueryOptions.queryOptions();
+		QueryOptions options = QueryOptions.queryOptions();
+		if (options == null) { // add/override what we got from PseudoArgs
+			options = QueryOptions.queryOptions();
+		}
 		if (getParameters() != null) {
 			if (getParameters() instanceof JsonArray) {
 				options.parameters((JsonArray) getParameters());
@@ -331,10 +334,13 @@ public class Query {
 				options.parameters((JsonObject) getParameters());
 			}
 		}
+		if (scanConsistency == null
+				|| scanConsistency == QueryScanConsistency.NOT_BOUNDED && getScanConsistency() != null) {
+			scanConsistency = getScanConsistency();
+		}
 		if (scanConsistency != null) {
 			options.scanConsistency(scanConsistency);
 		}
-
 		return options;
 	}
 
--- a/src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
@@ -73,4 +73,16 @@ public class StringQuery extends Query {
 		appendSkipAndLimit(statement);
 		return statement.toString();
 	}
+
+	/**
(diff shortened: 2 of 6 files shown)


src/main/java/org/springframework/data/couchbase/core/query/Query.java
```diff
@@ -323,18 +323,24 @@ public String toN1qlRemoveString(ReactiveCouchbaseTemplate template, String coll
 	 * @return QueryOptions
 	 */
 	public QueryOptions buildQueryOptions(QueryScanConsistency scanConsistency) {
-		final QueryOptions options = QueryOptions.queryOptions();
+		QueryOptions options = QueryOptions.queryOptions();
+		if (options == null) { // add/override what we got from PseudoArgs
```
daschl, line 327: that line does not make sense, since the line above you always set it to non-null?

src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
```diff
@@ -73,4 +73,16 @@ public String toN1qlSelectString(ReactiveCouchbaseTemplate template, String coll
 		appendSkipAndLimit(statement);
 		return statement.toString();
 	}
+
+	@Override
```
daschl, line 77: move below block comment?

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/core/query/Query.java
mikereiche: Right.  The code is copied from the Collections branch to reduce the future merge.   And in that code 'options' is an argument and needs to be checked.

src/main/java/org/springframework/data/couchbase/core/query/StringQuery.java
mikereiche: Yes. Will fix.

PR #1142, opened 2021-05-28
Add missing hooks for callbacks.

### LINKED ISSUE #1074: AfterSaveEvent is never triggered
`AbstractCouchbaseEventListener#onAfterSave` is never called

This worked in Spring Data Couchbase 3.x

mikereiche:
1) the previous change (#955) to CouchbaseAuditingRegister intended to register
 AuditingEntityCallback instead of AuditingEventListener but only changed the
name that AuditingEventListener was registered with. Since
AuditingEventListener does exactly the same thing as AuditingEntityCallback,
it behaved as those AuditingEntityCallback was being used.

2) It seems that the two mechanisms - EntityCallback and EventListener
both do more-or-less the same thing. I'm not sure why there are two
mechanisms to do the same thing.

3) Although the ReactiveAuditingEntityCallback mechanism is triggered
only for reactive operations and the non-reactive AuditingEntityCallback
is triggered only for non-reactive operations, the events are triggered
for both. Since there was a desire to have distinct callbacks for
reactive/non-reactive operations, I also made distinct events for
reactive/non-reactive operations.  However both the reactive and non-reactive
event mechanisms are triggered for either reactive and non-reactive operations,
so it is necessary for the mechanism to determine if it should act on the
event.

4) Although there is an AbstractCouchbaseEventListener that is extended by
LoggingEventListener and ValidatingEventListener, AuditingEventListener does
not extend it (maybe it should).

Closes #1074.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
@@ -26,6 +26,7 @@ import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
 import org.springframework.data.couchbase.core.mapping.event.AfterConvertCallback;
+import org.springframework.data.couchbase.core.mapping.event.AfterSaveEvent;
 import org.springframework.data.couchbase.core.mapping.event.BeforeConvertCallback;
 import org.springframework.data.couchbase.core.mapping.event.BeforeConvertEvent;
 import org.springframework.data.couchbase.core.mapping.event.BeforeSaveEvent;
@@ -95,16 +96,21 @@ class CouchbaseTemplateSupport implements ApplicationContextAware, TemplateSuppo
 	}
 
 	@Override
-	public Object applyUpdatedCas(final Object entity, final long cas) {
+	public Object applyUpdatedCas(final Object entity, CouchbaseDocument converted, final long cas) {
+		Object returnValue;
 		final ConvertingPropertyAccessor<Object> accessor = getPropertyAccessor(entity);
 		final CouchbasePersistentEntity<?> persistentEntity = mappingContext.getRequiredPersistentEntity(entity.getClass());
 		final CouchbasePersistentProperty versionProperty = persistentEntity.getVersionProperty();
 
 		if (versionProperty != null) {
 			accessor.setProperty(versionProperty, cas);
-			return accessor.getBean();
+			returnValue = accessor.getBean();
+		} else {
+			returnValue = entity;
 		}
-		return entity;
+		maybeEmitEvent(new AfterSaveEvent(returnValue, converted));
+
+		return returnValue;
 	}
 
 	@Override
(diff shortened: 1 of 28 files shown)


src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -172,7 +178,7 @@ public void setEntityCallbacks(EntityCallbacks entityCallbacks) {
 		this.entityCallbacks = entityCallbacks;
 	}
 
-	void maybeEmitEvent(CouchbaseMappingEvent<?> event) {
+	public void maybeEmitEvent(CouchbaseMappingEvent<?> event) {
 		if (canPublishEvent()) {
```
mikereiche, line 182: need to call from ReactiveRemoveByIdOperationSupport

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -95,16 +96,21 @@ public CouchbaseDocument encodeEntity(final Object entityToEncode) {
 	}
 
 	@Override
-	public Object applyUpdatedCas(final Object entity, final long cas) {
+	public Object applyUpdatedCas(final Object entity, CouchbaseDocument converted, final long cas) {
```
mikereiche, line 99: Added converted CouchbaseDocument to args as AfterSaveEvent takes a CouchbaseDocument.

src/main/java/org/springframework/data/couchbase/core/mapping/event/AfterConvertCallback.java
```diff
@@ -15,29 +15,27 @@
...
+	 * either the same or a modified instance of the domain object.
 	 *
 	 * @param entity the domain object (the result of the conversion).
 	 * @param document must not be {@literal null}.
 	 * @param collection name of the collection.
-	 * @return the domain object that is the result of reading it from the {@link Document}.
+	 * @return the domain object that is the result of reading it from the {@link CouchbaseDocument}.
 	 */
-	T onAfterConvert(T entity, Document document, String collection);
+	T onAfterConvert(T entity, CouchbaseDocument document, String collection);
```
mikereiche, line 40: This was copied incorrectly from Mongo with the second arg of Document instead of CouchbaseDocument.

src/main/java/org/springframework/data/couchbase/core/mapping/event/AuditingEventListener.java
```diff
@@ -33,17 +35,19 @@
  * @author Mark Paluch
  * @author Michael Reiche
  */
-public class AuditingEventListener implements ApplicationListener<BeforeConvertEvent<Object>> {
+public class AuditingEventListener implements ApplicationListener<CouchbaseMappingEvent<Object>> {
```
mikereiche, line 38: Support any CouchbaseMappingEvent.  Not just BeforeConvertEvent.

src/main/java/org/springframework/data/couchbase/core/mapping/event/ReactiveAuditingEntityCallback.java
```diff
@@ -57,9 +63,22 @@ public ReactiveAuditingEntityCallback(final ObjectFactory<ReactiveIsNewAwareAudi
...
+	/*
+	 * (non-Javadoc)
+	 *
+	 * @see
+	 * org.springframework.data.couchbase.core.mapping.event.ReactiveAfterConvertCallback#onAfterConvert
+	 * (java.lang.Object, CouchbaseDocument, java.lang.String)
+	 */
+	@Override
+	public Publisher<Object> onAfterConvert(Object entity, CouchbaseDocument document, String collection) {
+		LOG.debug("onAfterConvert " + document.toString());
```
mikereiche, line 79: should remove this debug.

PR #1148, opened 2021-06-08
Improve framework for custom mapping couchbase converter.

### LINKED ISSUE #1141: Custom converter for multibuckets
i implemented multibuckets according to explained in [this issue](https://github.com/spring-projects/spring-data-couchbase/issues/878) but i can't add custom converters to all buckets, the converters added for default bucket only

please see my code:

Custome converter
```
 @Override
   public CouchbaseCustomConversions customConversions() {
       return new CouchbaseCustomConversions(Arrays.asList(
               ZonedDateTimeToEpochTimeConverter.INSTANCE,
               EpochTimeToZonedDateTimeConverter.INSTANCE));
   }


   @WritingConverter
   public enum ZonedDateTimeToEpochTimeConverter implements Converter<ZonedDateTime, CouchbaseDocument> {
       INSTANCE;

       @Override
       public CouchbaseDocument convert(ZonedDateTime zonedDateTime) {
           CouchbaseDocument cd = new CouchbaseDocument();
           cd.put(EPOC_MILLI, zonedDateTime.toInstant().toEpochMilli());
           cd.put(OFFSET_SECONDS, zonedDateTime.getOffset().getTotalSeconds());
           return cd;
       }
   }

   @ReadingConverter
   public enum EpochTimeToZonedDateTimeConverter implements Converter<CouchbaseDocument, ZonedDateTime> {
       INSTANCE;

       @Override
       public ZonedDateTime convert(CouchbaseDocument epochTime) {
           long timeMilli = Long.parseLong(epochTime.getContent().get(EPOC_MILLI).toString());
           int offsetSeconds = Integer.parseInt(epochTime.getContent().get(OFFSET_SECONDS).toString());

           ZoneOffset convertedOffset = ZoneOffset.ofTotalSeconds(offsetSeconds);
           return ZonedDateTime.ofInstant(Instant.ofEpochMilli(timeMilli), convertedOffset);
       }
   }
```
add converter to the costum mapping:
```

MappingCouchbaseConverter mappingCouchbaseConverter = new MappingCouchbaseConverter();
mappingCouchbaseConverter.setCustomConversions(customConversions());
baseMapping.mapEntity(Class.forName(beanDefinition.getBeanClassName()),myCouchbaseTemplate(myCouchbaseClientFactory("mybucket"),mappingCouchbaseConverter ));

```
I am getting below exception when save the doc although there is convertor from zonedatetime to couchebasedocument
```

2021-05-27 17:29:27.178 ERROR 18916 --- [nio-9001-exec-1] g.k.e.error.DefaultGraphQLErrorHandler   : Error executing query Exception while fetching data (/createProduct) : No converter found capable of converting from type [java.time.ZonedDateTime] to type [org.springframework.data.couchbase.core.mapping.CouchbaseDocument]

org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [java.time.ZonedDateTime] to type [org.springframework.data.couchbase.core.mapping.CouchbaseDocument]
   at org.springframework.core.convert.support.GenericConversionService.handleConverterNotFound(GenericConversionService.java:322) ~[spring-core-5.3.5.jar:5.3.5]
   at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:195) ~[spring-core-5.3.5.jar:5.3.5]
   at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:175) ~[spring-core-5.3.5.jar:5.3.5]
   at org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter.lambda$getPotentiallyConvertedSimpleWrite$2(MappingCouchbaseConverter.java:784) ~[spring-data-couchbase-4.1.6.jar:4.1.6]
   at java.base/java.util.Optional.map(Optional.java:265) ~[na:na]
   at org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter.getPotentiallyConvertedSimpleWrite(MappingCouchbaseConverter.java:784) ~[spring-data-couchbase-4.1.6.jar:4.1.6]
   at org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter.writeSimpleInternal(MappingCouchbaseConverter.java:774) ~[spring-data-couchbase-4.1.6.jar:4.1.6]
   at org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter.access$200(MappingCouchbaseConverter.java:82) ~[spring-data-couchbase-4.1.6.jar:4.1.6]
```
   
Can you please help?

mikereiche:
1) change examples in Config to show creating repository mappings using
existing MappingCouchbaseConverter (which have customConversions /
BeanNames.COUCHBASE_CUSTOM_CONVERSIONS and applicationContext)
instead of creating their own and then adding the customConversion and
applicationContext.

2) remove unused members fro StringBasedN1qlQueryParser

Closes #1141.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/AbstractCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/AbstractCouchbaseConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors
+ * Copyright 2012-2021 the original author or authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@
 package org.springframework.data.couchbase.core.convert;
 
 import java.util.Collections;
-import java.util.Optional;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.core.convert.ConversionService;
@@ -68,7 +67,8 @@ public abstract class AbstractCouchbaseConverter implements CouchbaseConverter, 
 	}
 
 	/**
-	 * Set the custom conversions.
+	 * Set the custom conversions. Note that updating conversions requires a subsequent call to register them with the
+	 * conversionService: conversions.registerConvertersIn(conversionService)
 	 *
 	 * @param conversions the conversions.
 	 */
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors
+ * Copyright 2012-2021 the original author or authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,7 +49,6 @@ import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
 import com.couchbase.client.core.deps.com.fasterxml.jackson.databind.DeserializationFeature;
-import com.couchbase.client.core.deps.com.fasterxml.jackson.databind.Module;
 import com.couchbase.client.core.encryption.CryptoManager;
 import com.couchbase.client.core.env.Authenticator;
 import com.couchbase.client.core.env.PasswordAuthenticator;
@@ -60,7 +59,6 @@ import com.couchbase.client.java.encryption.databind.jackson.EncryptionModule;
 import com.couchbase.client.java.env.ClusterEnvironment;
 import com.couchbase.client.java.json.JacksonTransformers;
 import com.couchbase.client.java.json.JsonValueModule;
-import com.couchbase.client.java.json.RepackagedJsonValueModule;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
@@ -74,8 +72,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
  */
 @Configuration
 public abstract class AbstractCouchbaseConfiguration {
-
-	@Autowired ObjectMapper couchbaseObjectMapper;
 
 	/**
 	 * The connection string which allows the SDK to connect to the cluster.
@@ -144,7 +140,7 @@ public abstract class AbstractCouchbaseConfiguration {
 		if (!nonShadowedJacksonPresent()) {
 			throw new CouchbaseException("non-shadowed Jackson not present");
 		}
-		builder.jsonSerializer(JacksonJsonSerializer.create(couchbaseObjectMapper));
+		builder.jsonSerializer(JacksonJsonSerializer.create(couchbaseObjectMapper()));
 		configureEnvironment(builder);
 		return builder.build();
 	}
--- a/src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2020 the original author or authors.
+ * Copyright 2017-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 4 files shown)


src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -148,6 +148,8 @@ public MappingCouchbaseConverter(
 		this.mappingContext = mappingContext;
 		typeMapper = new DefaultCouchbaseTypeMapper(typeKey != null ? typeKey : TYPEKEY_DEFAULT);
 		spELContext = new SpELContext(CouchbaseDocumentPropertyAccessor.INSTANCE);
+		// ApplicationContext is needed for environment for resolving SPEL expressions.
+		this.setApplicationContext(((CouchbaseMappingContext) mappingContext).getApplicationContext());
```
daschl, line 152: should we throw an exception if the mappingContext is not instanceof CouchbaseMappingContext? So it will be descriptive

daschl (approved):
+2, just one minor comment that can slightly improve user experience. but feel free to merge

src/main/java/org/springframework/data/couchbase/core/convert/AbstractCouchbaseConverter.java
```diff
@@ -74,6 +73,7 @@ public ConversionService getConversionService() {
 	 */
 	public void setCustomConversions(final CustomConversions conversions) {
 		this.conversions = conversions;
+		afterPropertiesSet(); // register the conversions with the conversion service
```
mp911de, line 76: Please refrain from calling lifecycle methods in setters. Calling `afterPropertiesSet` is a responsibility of the bean container.

mikereiche: I can fix that by instead calling conversions.registerConvertersIn(conversionService) directly here.

The underlying issue is that 
(a) setCustomConversions() doesn't do everything needed to setCustomConversions(). It only does this.conversions = conversions and that isn't sufficient. It is also necessary to conversions.registerConvertersIn(conversionService) - otherwise the conversionService is not aware of the conversions. I made the call to afterPropertiesSet() in the event that in the future additional actions were added.
(b) the mapping converter was created outside the Bean container - in #1141 - although that should be unnecessary by  creating the template for configureRepositoryOperationsMapping using the converter (MappingCouchbaseConverter) (baseMapping.getDefault().getConverter()) . 

Somewhat related - Intellij complains when an @Bean is called directly, but sometimes it is necessary - when the value of an @Bean method is being used in another public @Bean, and the signature of that public @Bean cannot (should not?) be changed.  This is the situation with  On the other hand - if there was a guarantee that no applications were calling the @Bean method directly, then changing the signature should not be a problem. This issue exists in three bean methods in AbstractCouchbaseConfiguration.   The "solution" has been to make @Bean method call - the problem with this is that the result of that call is not the actual @Bean, it's a different object. (I suppose that if the result object was cached and used on subsequent calls, then it could be the same object).

couchbaseObjectMapper() is an @Bean method, it is now required by the couchbaseClusterEnvironment() @Bean, but was not/is not a method argument
```
	@Bean(destroyMethod = "shutdown")
	public ClusterEnvironment couchbaseClusterEnvironment() 
	   ... 
	   builder.jsonSerializer(JacksonJsonSerializer.create(couchbaseObjectMapper()));
```

This couchbaseTemplate() method was previously the @Bean, but has been superseded with a method with an additional argument.   
```
	public CouchbaseTemplate couchbaseTemplate(CouchbaseClientFactory couchbaseClientFactory,
			MappingCouchbaseConverter mappingCouchbaseConverter) {
		return couchbaseTemplate(couchbaseClientFactory, mappingCouchbaseConverter, new JacksonTranslationService());
	}
```

mp911de: > setCustomConversions() doesn't do everything needed to setCustomConversions(). It only doe

That's by intent. The bean configuration phase consists from calling all setters first and then initialize the bean with `afterPropertiesSet`. That being said, calling `registerConvertersIn(…)` in the setter isn't better.

>  the mapping converter was created outside the Bean container 

It is then the responsibility of the creator to initialize that bean.

mikereiche: ok. I'll just remove the call and not replace it with anything.

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
mp911de: Actually, the converter should be ApplicationContextAware so that the bean container injects the context. You could skip that line and stop exposing the context from the mapping context for a clean implementation.

mikereiche: Yes.  This shouldn't be done here. It's similar to afterPropertiesSet() call, in that it should normally be done by the Bean container. And if the object is created outside the bean container, then the caller has to take care of the initialization that the bean container would normally do.   So I'll remove this and not expose the applicationContext of the mapping context.

PR #1149, opened 2021-06-17
Scopes and collections for repositories

### LINKED ISSUE #963: Support for Scope and Collection API Changes [DATACOUCH-655]
**[Arun Vijayraghavan](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER49896)** opened **[DATACOUCH-655](https://jira.spring.io/browse/DATACOUCH-655?redirect=false)** and commented

Not sure if this ticket is needed, but this will stand as a place holder for the API changes w.r.t underlying server specific scopes and collections API Changes.

 

Please feel free to close this ticket if you think this will be taken care by the JAVA SDK

---
No further details from [DATACOUCH-655](https://jira.spring.io/browse/DATACOUCH-655?redirect=false)

mikereiche:
This adds scope and collection support to repositories via the
DynamicInvocationHandler and PseudoArgs.
It also adds annotations for scopes and collections.

Closes #963

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/SimpleCouchbaseClientFactory.java
+++ b/src/main/java/org/springframework/data/couchbase/SimpleCouchbaseClientFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors
+ * Copyright 2012-2021 the original author or authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.springframework.data.couchbase;
 
 import java.util.function.Supplier;
@@ -33,6 +32,9 @@ import com.couchbase.client.java.env.ClusterEnvironment;
 
 /**
  * The default implementation of a {@link CouchbaseClientFactory}.
+ *
+ * @author Michael Nitschinger
+ * @author Michael Reiche
  */
 public class SimpleCouchbaseClientFactory implements CouchbaseClientFactory {
 
@@ -74,7 +76,7 @@ public class SimpleCouchbaseClientFactory implements CouchbaseClientFactory {
 
 	@Override
 	public CouchbaseClientFactory withScope(final String scopeName) {
-		return new SimpleCouchbaseClientFactory(cluster, bucket.name(), scopeName);
+		return new SimpleCouchbaseClientFactory(cluster, bucket.name(), scopeName != null ? scopeName : getScope().name());
 	}
 
 	@Override
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
@@ -106,6 +106,7 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 	}
 
 	@Override
+	@Deprecated
 	public ExecutableRemoveById removeById() {
 		return removeById(null);
 	}
@@ -116,6 +117,7 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 	}
 
 	@Override
+	@Deprecated
 	public ExecutableExistsById existsById() {
 		return existsById(null);
 	}
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
@@ -28,7 +28,6 @@ import org.springframework.data.couchbase.core.index.CouchbasePersistentEntityIn
 import org.springframework.data.couchbase.core.mapping.CouchbaseMappingContext;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentEntity;
 import org.springframework.data.couchbase.core.mapping.CouchbasePersistentProperty;
-import org.springframework.data.couchbase.core.support.PseudoArgs;
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.lang.Nullable;
 
@@ -108,12 +107,22 @@ public class CouchbaseTemplate implements CouchbaseOperations, ApplicationContex
 
 	@Override
 	public ExecutableRemoveById removeById() {
-		return new ExecutableRemoveByIdOperationSupport(this).removeById();
+		return removeById(null);
+	}
+
+	@Override
+	public ExecutableRemoveById removeById(Class<?> domainType) {
+		return new ExecutableRemoveByIdOperationSupport(this).removeById(domainType);
 	}
 
 	@Override
 	public ExecutableExistsById existsById() {
-		return new ExecutableExistsByIdOperationSupport(this).existsById();
+		return existsById(null);
+	}
+
+	@Override
+	public ExecutableExistsById existsById(Class<?> domainType) {
+		return new ExecutableExistsByIdOperationSupport(this).existsById(domainType);
 	}
 
 	@Override
--- a/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
@@ -34,6 +34,7 @@ import org.springframework.context.ApplicationContextAware;
 import org.springframework.core.CollectionFactory;
 import org.springframework.core.convert.ConversionService;
 import org.springframework.core.convert.support.DefaultConversionService;
+import org.springframework.data.annotation.Transient;
 import org.springframework.data.convert.EntityInstantiator;
 import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
 import org.springframework.data.couchbase.core.mapping.CouchbaseList;
@@ -530,6 +531,10 @@ public class MappingCouchbaseConverter extends AbstractCouchbaseConverter implem
 						IdAttribute idAttribute = prop.findAnnotation(IdAttribute.class);
 						int order = idAttribute.order();
 						idAttributes.put(order, convertToString(propertyObj));
+					}
+
+					if (prop.isAnnotationPresent(Transient.class)) {
+						return;
 					}
 
 					if (!conversions.isSimpleType(propertyObj.getClass())) {
--- a/src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
+++ b/src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
@@ -21,10 +21,9 @@ import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
 import org.springframework.context.EnvironmentAware;
+import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.core.env.Environment;
-import org.springframework.data.annotation.Id;
 import org.springframework.data.mapping.MappingException;
-import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.model.BasicPersistentEntity;
 import org.springframework.data.util.TypeInformation;
 import org.springframework.util.Assert;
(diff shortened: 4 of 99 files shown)


src/main/java/org/springframework/data/couchbase/repository/auditing/ReactiveCouchbaseAuditingRegistrar.java
```diff
@@ -82,17 +81,6 @@ protected void registerAuditListenerBeanDefinition(BeanDefinition auditingHandle
...
-		BeanDefinitionBuilder builder2 = BeanDefinitionBuilder.rootBeanDefinition(ReactiveAuditingEventListener.class);
-
-		builder2
-				.addConstructorArgValue(ParsingUtils.getObjectFactoryBeanDefinition(getAuditingHandlerBeanName(), registry));
-		builder.getRawBeanDefinition().setSource(auditingHandlerDefinition.getSource());
-
-		registerInfrastructureBeanWithId(builder2.getBeanDefinition(), ReactiveAuditingEventListener.class.getName(),
-				registry);
-
 	}
```
mikereiche, line 84: There should not be any Reactive*Event classes.

src/main/java/org/springframework/data/couchbase/repository/support/CouchbaseRepositoryBase.java
```diff
@@ -0,0 +1,130 @@
+package org.springframework.data.couchbase.repository.support;
```
mikereiche, line 1: Factor out common components of CouchbaseRepository and ReactiveCouchbaseRepository

src/main/java/org/springframework/data/couchbase/repository/support/CrudMethodMetadataPostProcessor.java
```diff
@@ -161,10 +164,10 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
 				try {
 					return invocation.proceed();
 				} finally {
-					TransactionSynchronizationManager.unbindResource(method);
+					// TransactionSynchronizationManager.unbindResource(method);
 				}
 			} finally {
-				currentInvocation.set(oldInvocation);
+				// currentInvocation.set(oldInvocation);
```
mikereiche, line 170: unbinding the method and resetting the currentInvocation is problematic for  reactive calls - where the bound metadata maybe accessed in the Reactive Flux *after* the invocation is complete. 
Furthermore - unbinding/resetting is not necessary as this all occurs in the same thread.

src/main/java/org/springframework/data/couchbase/repository/support/DynamicInvocationHandler.java
```diff
@@ -0,0 +1,120 @@
...
+		Object result;
+
+		try {
+			setThreadLocal();
+			result = theMethod.invoke(target, args);
+		} catch (InvocationTargetException ite) {
+			throw ite.getCause();
+		}
+		return result;
+	}
```
mikereiche, line 111: similar to CrudMethodMetadataPostProcessor - there is no need to reset the thread-local data after the call as this is in the same thread.

src/main/java/org/springframework/data/couchbase/repository/support/SimpleReactiveCouchbaseRepository.java
```diff
@@ -97,7 +91,7 @@ public SimpleReactiveCouchbaseRepository(CouchbaseEntityInformation<T, String> e
 	@Override
 	public <S extends T> Flux<S> saveAll(Iterable<S> entities) {
 		Assert.notNull(entities, "The given Iterable of entities must not be null!");
-		return Flux.fromIterable(entities).flatMap(this::save);
+		return Flux.fromIterable(entities).flatMap(e -> save(e));
 	}
```
mikereiche, line 95: This call - where one reactive method calls another reactive method - is what causes the problem when the methodMetadata and methodInvocation are unbound/reset.   The "unbind/unset" occurs *before* save(e) is called, therefore in save(e) the threadLocal metadata is not accessible. The suggestion provided by Mark was to capture the threadLocal metadata in the outer reactive method, then passing it as an arg to the inner reactive method (below). Which requires a new method with that signature. It *seems* like not unbinding/resetting is ok as all our tests pass.  So maybe we can leave it as it is without adding methods with extra args.   Another solution might be to change all the methods like this so there is not a reactive method calling a reactive method.

`public <S extends T> Flux<S> saveAll(Iterable<S> entities) {
		Assert.notNull(entities, "The given Iterable of entities must not be null!");
                 PseudoArgs pArgs = template.getPseudoArgs();
		return Flux.fromIterable(entities).flatMap(e -> save(e, pArgs));
	}
`

src/main/java/org/springframework/data/couchbase/SimpleCouchbaseClientFactory.java
```diff
@@ -74,7 +76,7 @@ private SimpleCouchbaseClientFactory(final Supplier<Cluster> cluster, final Stri
 
 	@Override
 	public CouchbaseClientFactory withScope(final String scopeName) {
-		return new SimpleCouchbaseClientFactory(cluster, bucket.name(), scopeName);
+		return new SimpleCouchbaseClientFactory(cluster, bucket.name(), scopeName != null ? scopeName : getScope().name());
```
daschl, line 79: if this is null, should this use the default scope? Or should we add a withDefaultScope?

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
```diff
@@ -107,13 +106,25 @@ public CouchbaseTemplate(final CouchbaseClientFactory clientFactory, final Couch
 	}
 
 	@Override
+	@Deprecated
 	public ExecutableRemoveById removeById() {
-		return new ExecutableRemoveByIdOperationSupport(this).removeById();
+		return removeById(null);
```
daschl, line 111: I assume the domainType can be null in the operation supports?

src/main/java/org/springframework/data/couchbase/core/mapping/Expiry.java
```diff
@@ -0,0 +1,65 @@
...
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.concurrent.TimeUnit;
+
+import org.springframework.data.annotation.Persistent;
+
+/**
+ * Identifies a domain object to be persisted to Couchbase.
```
daschl, line 29: copy past error on the description?

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -532,6 +533,10 @@ public void doWithPersistentProperty(final CouchbasePersistentProperty prop) {
 						idAttributes.put(order, convertToString(propertyObj));
 					}
 
+					if (prop.isAnnotationPresent(Transient.class)) {
```
daschl, line 536: what does this do?

src/main/java/org/springframework/data/couchbase/repository/CouchbaseRepository.java
```diff
@@ -43,4 +47,8 @@
 	@Override
 	List<T> findAllById(Iterable<ID> iterable);
 
+	CouchbaseEntityInformation<T, String> getEntityInformation();
```
daschl, line 50: is it ok to add this here @mp911de ?

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplate.java
mikereiche: Yes. If provided, the domainType would allow accessing annotations on the entity class (such as the scope). Without, the scope would be determined by other means (possibly an annotation from the repository interface, or the scope of the clientFactory.

src/main/java/org/springframework/data/couchbase/SimpleCouchbaseClientFactory.java
mikereiche: getScope() is the scope of the clientFactory - that comes from AbstractCouchbaseConfiguration.getScope() or an overriding method.

daschl (changes requested):
This is getting close -- a couple comments/nits, but the big one is still that the interfaces should not get the new public signatures I think (it's in the comments)

src/main/java/org/springframework/data/couchbase/core/mapping/BasicCouchbasePersistentEntity.java
```diff
@@ -98,11 +97,14 @@ protected CouchbasePersistentProperty returnPropertyIfBetterIdPropertyCandidateO
 
 	@Override
 	public int getExpiry() {
-		Document annotation = getType().getAnnotation(Document.class);
+		return getExpiry(AnnotatedElementUtils.findMergedAnnotation(getType(), Expiry.class), environment);
+	}
+
+	public static int getExpiry(Expiry annotation, Environment environment) {
```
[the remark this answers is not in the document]
mikereiche: in MappingCouchbaseConverter, which is in core/convert

		target.setExpiration(entity.getExpiry());

src/main/java/org/springframework/data/couchbase/core/query/Meta.java
```diff
@@ -78,10 +85,26 @@ public boolean hasValues() {
 	 * @param key must not be {@literal null} or empty.
 	 * @param value
 	 */
-	void setValue(String key, @Nullable Object value) {
+	public void setValue(String key, @Nullable Object value) {
```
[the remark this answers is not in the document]
mikereiche: I don't think there is the possibility of access from multiple threads.

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -0,0 +1,329 @@
+package org.springframework.data.couchbase.core.query;
```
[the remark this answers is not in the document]
mikereiche: done

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -0,0 +1,329 @@
...
+		return options;
+	}
+
+	/**
+	 * scope annotation could be a
+	 * 
+	 * @param domainType
+	 * @return
+	 */
+	public static String getScopeFrom(Class<?> domainType) {
```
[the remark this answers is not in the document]
mikereiche: unfortunately, yes.  It's used by PseudoArgs which is in the core/support package.

src/main/java/org/springframework/data/couchbase/repository/CouchbaseRepository.java
mikereiche: Yes. I will move it to SimpleCouchbaseRepository and SimpleReactiveCouchbaseRepository.
(oddly, there is not option for me to add a comment here, but I can edit yours).

Edit: I will leave this here - I tried moving it to SimpleCouchbaseRepository, but casting the 'target' in DynamicInvocationHandler to a SimpleCouchbaseRepository results in a ClassCastException due to being loaded in a different class loader.

src/main/java/org/springframework/data/couchbase/repository/query/AbstractReactiveCouchbaseQuery.java
```diff
@@ -121,6 +116,8 @@ private ReactiveCouchbaseQueryExecution getExecutionToWrap(ParameterAccessor acc
 			return (q, t, c) -> operation.matching(q.with(accessor.getPageable())).all(); // s/b tail() instead of all()
 		} else if (getQueryMethod().isCollectionQuery()) {
 			return (q, t, c) -> operation.matching(q.with(accessor.getPageable())).all();
+			// } else if (getQueryMethod().isStreamQuery()) {
```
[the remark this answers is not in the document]
mikereiche: A reminder to give this attention.
Edit:  Stream<Entity> queries in Reactive are probably not necessary.

daschl (approved):

PR #1187, opened 2021-08-13
Log exceptions thrown from events.

### LINKED ISSUE #1075: CouchbaseTemplateSupport does not log exceptions thrown by CouchbaseMappingEvent Listeners correctly
`CouchbaseTemplateSupport#maybeEmitEvent` and `CouchbaseTemplateSupport#maybeCallBeforeConvert`

both contain

```java
			} catch (Exception e) {
				e.printStackTrace();
			}
```

mikereiche:
Closes #1075.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
@@ -182,7 +182,7 @@ class CouchbaseTemplateSupport implements ApplicationContextAware, TemplateSuppo
 			try {
 				this.applicationContext.publishEvent(event);
 			} catch (Exception e) {
-				e.printStackTrace();
+				LOG.error("exception emitting event "+event, e);
 			}
 		} else {
 			LOG.info("maybeEmitEvent called, but CouchbaseTemplate not initialized with applicationContext");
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/couchbase/core/CouchbaseTemplateSupport.java
```diff
@@ -182,7 +182,7 @@ public void maybeEmitEvent(CouchbaseMappingEvent<?> event) {
 			try {
 				this.applicationContext.publishEvent(event);
 			} catch (Exception e) {
-				e.printStackTrace();
+				LOG.error("exception emitting event", e);
```
daschl, line 185: you could add the event to the log message so in logging you can figure out which event caused the error

src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
```diff
@@ -189,7 +189,7 @@ public void maybeEmitEvent(CouchbaseMappingEvent<?> event) {
 			try {
 				this.applicationContext.publishEvent(event);
 			} catch (Exception e) {
-				e.printStackTrace();
+				LOG.error("exception emitting event", e);
```
daschl, line 192: you could add the event to the log message so in logging you can figure out which event caused the error

daschl (changes requested):

PR #1211, opened 2021-09-08
Support N1QL Join Annotations.

### LINKED ISSUE #1198: implement @N1qlJoin in spring-data-couchbase 4

mikereiche:
Closes #1198.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
@@ -18,8 +18,6 @@ package org.springframework.data.couchbase.core;
 
 import reactor.core.publisher.Mono;
 
-import java.util.List;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.BeansException;
@@ -37,15 +35,12 @@ import org.springframework.data.couchbase.core.mapping.event.BeforeSaveEvent;
 import org.springframework.data.couchbase.core.mapping.event.CouchbaseMappingEvent;
 import org.springframework.data.couchbase.core.mapping.event.ReactiveAfterConvertCallback;
 import org.springframework.data.couchbase.core.mapping.event.ReactiveBeforeConvertCallback;
-import org.springframework.data.couchbase.core.query.N1qlJoin;
 import org.springframework.data.couchbase.repository.support.MappingCouchbaseEntityInformation;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
-import org.springframework.data.mapping.PropertyHandler;
 import org.springframework.data.mapping.callback.EntityCallbacks;
 import org.springframework.data.mapping.callback.ReactiveEntityCallbacks;
 import org.springframework.data.mapping.context.MappingContext;
 import org.springframework.data.mapping.model.ConvertingPropertyAccessor;
-import org.springframework.data.util.TypeInformation;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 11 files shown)


src/test/java/org/springframework/data/couchbase/domain/Address.java
```diff
@@ -26,4 +44,20 @@ public void setCity(String city) {
 		this.city = city;
 	}
 
+	public String getParent_id() {
```
daschl, line 47: why mix underscore and camel case in that file?

src/main/java/org/springframework/data/couchbase/core/query/N1QLQuery.java
```diff
@@ -45,4 +47,21 @@ public JsonObject n1ql() {
 		return query;
 	}
 
+	@Override
+	public String toN1qlSelectString(ReactiveCouchbaseTemplate template, String collectionName, Class domainClass,
+			Class returnClass, boolean isCount, String[] distinctFields) {
+		/*
```
daschl, line 53: this big block commented out is odd - why?

src/main/java/org/springframework/data/couchbase/core/convert/join/N1qlJoinResolver.java
```diff
@@ -94,38 +109,63 @@ public static String buildQuery(CouchbaseTemplate template, N1qlJoinResolverPara
...
 
 	static public class N1qlJoinProxy implements InvocationHandler {
-		private final CouchbaseTemplate template;
+		private final Object template;
+		private final String collectionName = null;
 		private final N1qlJoinResolverParameters params;
 		private List<?> resolved = null;
 
-		public N1qlJoinProxy(CouchbaseTemplate template, N1qlJoinResolverParameters params) {
+		public N1qlJoinProxy(Object template, N1qlJoinResolverParameters params) {
```
daschl, line 150: the other option is to make the constructor accept both the regular and the reactive one and the caller just nulls out the one it doesn't use.. that way it still keeps the type safety.

src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
```diff
@@ -92,6 +100,22 @@ public ReactiveCouchbaseTemplateSupport(final CouchbaseConverter converter,
 			if (persistentEntity.getVersionProperty() != null) {
 				accessor.setProperty(persistentEntity.getVersionProperty(), cas);
 			}
+			persistentEntity.doWithProperties((PropertyHandler<CouchbasePersistentProperty>) prop -> {
```
daschl, line 103: if I'm not mistaken both these blocks on the reactive and non reactive one are identical? If so you can have it on one and call it from the other as a utility method to avoid the code duplication

daschl (changes requested):
As a side comment, I wonder if we should introduce a QueryJoin annotation which behaves exactly like n1qljoin (and deprecate that one) since it would align better with the naming in the framework and out other annotations?

src/test/java/org/springframework/data/couchbase/domain/Address.java
mikereiche: Fixed.

src/main/java/org/springframework/data/couchbase/core/query/N1QLQuery.java
mikereiche: cut-and-paste but not used.
Fixed.

src/main/java/org/springframework/data/couchbase/core/convert/join/N1qlJoinResolver.java
mikereiche: I made two constructors. Fixed.

src/main/java/org/springframework/data/couchbase/core/ReactiveCouchbaseTemplateSupport.java
mikereiche: Fixed.

PR #1212, opened 2021-09-08
Add project(fields) to findByQuery similar to same method on findById.

### LINKED ISSUE #1208: Support `project()` method for CouchbaseTemplate.findByQuery() queries
Currently, CouchbaseTemplate.findById() has a method project( fieldnames array ) that allows filter fields to project for a findById query. 

**_Request_**
Add findByQuery().project() method for projecting only a given list of fields.

mikereiche:
Closes #1208.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
@@ -50,7 +50,7 @@ public class ReactiveFindByQueryOperationSupport implements ReactiveFindByQueryO
 	@Override
 	public <T> ReactiveFindByQuery<T> findByQuery(final Class<T> domainType) {
 		return new ReactiveFindByQuerySupport<>(template, domainType, domainType, ALL_QUERY, null, null, null, null, null,
-				template.support());
+				null, template.support());
 	}
 
 	static class ReactiveFindByQuerySupport<T> implements ReactiveFindByQuery<T> {
@@ -63,12 +63,13 @@ public class ReactiveFindByQueryOperationSupport implements ReactiveFindByQueryO
 		private final String collection;
 		private final String scope;
 		private final String[] distinctFields;
+		private final String[] fields;
 		private final QueryOptions options;
 		private final ReactiveTemplateSupport support;
 
 		ReactiveFindByQuerySupport(final ReactiveCouchbaseTemplate template, final Class<?> domainType,
 				final Class<T> returnType, final Query query, final QueryScanConsistency scanConsistency, final String scope,
-				final String collection, final QueryOptions options, final String[] distinctFields,
+				final String collection, final QueryOptions options, final String[] distinctFields, final String[] fields,
 				final ReactiveTemplateSupport support) {
 			Assert.notNull(domainType, "domainType must not be null!");
 			Assert.notNull(returnType, "returnType must not be null!");
@@ -81,6 +82,7 @@ public class ReactiveFindByQueryOperationSupport implements ReactiveFindByQueryO
 			this.collection = collection;
 			this.options = options;
 			this.distinctFields = distinctFields;
+			this.fields = fields;
 			this.support = support;
 		}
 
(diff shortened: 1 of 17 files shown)


src/main/java/org/springframework/data/couchbase/core/support/WithProjecting.java
```diff
@@ -0,0 +1,27 @@
+/*
```
daschl, line 1: is that interface actually used somwhere in this pr?

src/main/java/org/springframework/data/couchbase/core/support/WithProjecting.java
```diff
@@ -0,0 +1,27 @@
...
+ * Copyright 2020 the original author or authors
```
daschl, line 2: outdated date

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -177,50 +182,64 @@ private String getDistinctFields(String... distinctFields) {
...
+					if (prop.isVersionProperty()) {
+						return;
+					}
+					if (fieldsList != null && !fieldsList.contains(prop.getFieldName())) {
+						return;
+					}
+					PersistentPropertyPath<CouchbasePersistentProperty> path = couchbaseConverter.getMappingContext()
+							.getPersistentPropertyPath(prop.getName(), resultClass.getType());
+
+					// The current limitation is that only top-level properties can be projected
```
daschl, line 219: can we detect this and throw an error if so?

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -177,50 +182,64 @@ private String getDistinctFields(String... distinctFields) {
...
+					// This traversing of nested data structures would need to replicate the processing done by
+					// MappingCouchbaseConverter. Either the read or write
+					// And the n1ql to project lower-level properties is complex
+
+					// if (!conversions.isSimpleType(prop.getType())) {
+					// getProjectedFieldsInternal(prop, sb, prop.getTypeInformation(), path+prop.getName()+".");
+					// } else {
+					if (sb.length() > 0) {
+						sb.append(", ");
```
daschl, line 228: is a trailing , trimmed somewhere?

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -127,28 +129,31 @@ public StringBasedN1qlQueryParser(String statement, CouchbaseQueryMethod queryMe
...
+		this.countContext = createN1qlSpelValues(bucketName, collection, null, null, typeField, typeValue, true, null,
+				null);
 		this.parsedExpression = getExpression(accessor, getParameters(accessor), null, parser, evaluationContextProvider);
 		checkPlaceholders(this.parsedExpression.toString());
 	}
 
 	public StringBasedN1qlQueryParser(String bucketName, String collection, CouchbaseConverter couchbaseConverter,
-			Class domainClass, Class resultClass, String typeField, String typeValue, String[] distinctFields) {
+			Class domainClass, Class resultClass, String typeField, String typeValue, String[] distinctFields,
+			String[] fields) {
```
daschl, line 142: what happens if both distinctFields and fields are provided? Is OK or do we need to fail on that?

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
```diff
@@ -94,46 +96,52 @@ public ReactiveFindByQueryOperationSupport(final ReactiveCouchbaseTemplate templ
...
 		public <R> FindByQueryWithConsistency<R> as(Class<R> returnType) {
 			Assert.notNull(returnType, "returnType must not be null!");
 			return new ReactiveFindByQuerySupport<>(template, domainType, returnType, query, scanConsistency, scope,
-					collection, options, distinctFields, support);
+					collection, options, distinctFields, fields, support);
+		}
+
+		@Override
+		public FindByQueryWithProjection<T> project(String[] fields) {
+			// Assert.notNull(fields, "Fields must not be null");
```
daschl, line 142: should we do a null check here?

daschl (changes requested):

src/main/java/org/springframework/data/couchbase/core/support/WithProjecting.java
mikereiche: It's not.  It was intended to be used on all classes that have this project() api, but I couldn't figure out how to resolve this project() method with the one being implemented.

src/main/java/org/springframework/data/couchbase/core/support/WithProjecting.java
mikereiche: fixed.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
mikereiche: Yes.  Fixed.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
mikereiche: No need - the comma is only output if there is already something output and something else is following.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
mikereiche: If distinctFields[] is non-null, then it will be a "distinct" query - and if distinctFields[] is non-empty, then those fields are DISTINCT().  If distinctFields is non-null, but empty, then DISTINCT() will contain all the project fields - which could be (a) all the fields in the repository entity, (b) all the fields in an alternate entity (via as(class)); or (c) all the projected fields[]. 

So having distinctFields[] non-null and empty with fields[] specifying the desired fields gives the same result as having distinctFields[] specifying the desired fields.  Maybe not allow the first?

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByQueryOperationSupport.java
mikereiche: Yes. And also in ExecutableFindByQueryOperationSupport.project(). Fixed.

PR #1241, opened 2021-09-24
Fix flaky test in org.springframework.data.couchbase.core.mapping.MappingCouchbaseConverterTests.writesAndReadsCustomConvertedClass

CharlesZKQ:
`org.springframework.data.couchbase.core.mapping.MappingCouchbaseConverterTests.writesAndReadsCustomConvertedClass` has a flaky test because it directly uses `.toString()` when wanted to compare the actual value in the `HashMap` for `source` and `converted`. Every time you use `.toString()` for `HashMap`, it might construct the string in a different order. I fixed it by get all elements in the `HashMap` and compare them one by one.


mikereiche (dismissed):

mikereiche:
Just curious - which JVM is this failing with?

CharlesZKQ:
> Just curious - which JVM is this failing with?

Thank you for you rely, I used the JDK 8 (I believe the JVM should be 8 as well)

CharlesZKQ:
I squashed 4 commits into 1, does that count commit message fix?

mikereiche:
This is being pulled in via a different PR.

PR #1253, opened 2021-10-18
fix flaky for writesAndReadsClassContainingCustomConvertedObjects

YujiaYan0709:
`org.springframework.data.couchbase.core.mapping.MappingCouchbaseConverterTests.writesAndReadsClassContainingCustomConvertedObjects` is a flaky test because it directly use `toString()` to convert the content of the HashMap into String, which is non-deterministic for the order of the key-value pair. I fixed it by instead, compare the HashMap directly by assert.


mikereiche:
This is being pulled in via a different PR.

PR #1256, opened 2021-10-18
Fix SimpleCouchbaseRepository not taking into account the default consistency.

### LINKED ISSUE #1243: Set default consistency at project level, per profile
According to the documentation, the default consistency level for Repositories is `NOT_BOUNDED` ( https://docs.spring.io/spring-data/couchbase/docs/current/reference/html/#couchbase.repository.consistency ).
This is fine for some environment, but not for others. Typically during tests, most of them fail because they try to read their own writes, and it fails.
With a previous version of Spring Data Couchbase based on SDK 2.x we could extend `getDefaultConsistency()` from `AbstractCouchbaseConfiguration`. This method is now gone and as we are migrating, it breaks our app. No green tests, no delivery.

JesusTheHun:
Setting default consistency should make all queries performed with said consistency. 

This PR fixes the issue and add a non-regression test.

Closes #1243 .

@pivotal-cla This is an Obvious Fix

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
@@ -16,30 +16,9 @@
 
 package org.springframework.data.couchbase.repository;
 
-import static com.couchbase.client.java.query.QueryScanConsistency.NOT_BOUNDED;
-import static com.couchbase.client.java.query.QueryScanConsistency.REQUEST_PLUS;
-import static java.util.Arrays.asList;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.springframework.data.couchbase.config.BeanNames.COUCHBASE_TEMPLATE;
-
-import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/couchbase/repository/support/SimpleCouchbaseRepository.java
```diff
@@ -190,7 +191,12 @@ public void deleteAll() {
 	}
 
 	private QueryScanConsistency buildQueryScanConsistency() {
-		QueryScanConsistency scanConsistency = QueryScanConsistency.NOT_BOUNDED;
+		QueryScanConsistency scanConsistency = null;
+
+		if (couchbaseOperations.getConsistency() != null) {
+			scanConsistency = couchbaseOperations.getConsistency();
+		}
+
```
mikereiche, line 199: Only the first line is necessary.  The Reactive*OperationSupport.build*Options() will take consistency from the template (couchbaseOperations) if not otherwise set.

src/test/java/org/springframework/data/couchbase/repository/CouchbaseRepositoryQueryIntegrationTests.java
```diff
@@ -57,20 +57,13 @@
...
-import org.springframework.data.couchbase.domain.PersonRepository;
-import org.springframework.data.couchbase.domain.User;
-import org.springframework.data.couchbase.domain.UserRepository;
+import org.springframework.data.couchbase.domain.*;
 import org.springframework.data.couchbase.domain.time.AuditingDateTimeProvider;
 import org.springframework.data.couchbase.repository.auditing.EnableCouchbaseAuditing;
 import org.springframework.data.couchbase.repository.config.EnableCouchbaseRepositories;
 import org.springframework.data.couchbase.repository.query.CouchbaseQueryMethod;
 import org.springframework.data.couchbase.repository.query.CouchbaseRepositoryQuery;
+import org.springframework.data.couchbase.repository.support.SimpleCouchbaseRepository;
```
mikereiche, line 66: Not needed?

PR #1279, opened 2021-12-08
Port to jdk17 and spring 3.0.0

### LINKED ISSUE #1278: Upgrade Java baseline to Java 17
jdk17 on 5.0.x branch

mikereiche:
Closes #1278.


pom.xml
```diff
@@ -280,15 +277,15 @@
...
-              <groupId>org.apache.maven.plugins</groupId>
-              <artifactId>maven-jar-plugin</artifactId>
-              <configuration>
-                <archive>
-                  <manifestEntries>
-                    <Automatic-Module-Name>${java-module-name}</Automatic-Module-Name>
-                  </manifestEntries>
-                </archive>
-              </configuration>
+                <groupId>org.apache.maven.plugins</groupId>
```
mp911de, line 280: Note that this is already configured by Spring Data Build Parent.

src/main/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreator.java
```diff
@@ -69,7 +72,13 @@ public N1qlQueryCreator(final PartTree tree, final ParameterAccessor accessor, f
 		this.queryMethod = queryMethod;
 		this.converter = converter;
 		this.bucketName = bucketName;
-		this.entity = converter.getMappingContext().getPersistentEntity(queryMethod.getReturnedObjectType());
+
+		Class<?> returnedObjectType = queryMethod.getReturnedObjectType();
+		if (ClassUtils.isPrimitiveOrWrapper(returnedObjectType)) {
+			this.entity = new BasicCouchbasePersistentEntity(ClassTypeInformation.from(returnedObjectType));
```
mp911de, line 78: Entities should not be generally created for return types. A query is always contextually related to an entity, but using the return type doesn't match that assumption. You could return `void` or `Integer` and entity creation inspects constructors etc. That could should be adopted properly to detect whether the return type qualifies as entity.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -194,7 +197,12 @@ private String getProjectedFields(String b, Class resultClass, String[] fields)
 
 		String projectedFields = i(b) + ".*";
 		if (resultClass != null) {
-			PersistentEntity persistentEntity = couchbaseConverter.getMappingContext().getPersistentEntity(resultClass);
+			PersistentEntity persistentEntity;
+			if (ClassUtils.isPrimitiveOrWrapper(resultClass)) {
+				persistentEntity = new BasicCouchbasePersistentEntity(ClassTypeInformation.from(resultClass));
```
mp911de, line 202: Similar issue as with `N1qlQueryCreator`. The entity should be determined on the query method level and if the return type qualifies as entity, the entity to be used should be refined into the type instead of creating `PersistentEntity` instances.

src/main/java/org/springframework/data/couchbase/repository/query/StringBasedN1qlQueryParser.java
```diff
@@ -207,7 +215,12 @@ private void getProjectedFieldsInternal(String bucketName, CouchbasePersistentPr
 
 		if (resultClass != null) {
 			Set<String> fieldList = fields != null ? new HashSet<>(Arrays.asList(fields)) : null;
-			PersistentEntity persistentEntity = couchbaseConverter.getMappingContext().getPersistentEntity(resultClass);
+			PersistentEntity persistentEntity;
+			if (ClassUtils.isPrimitiveOrWrapper(resultClass.getType())) {
```
mp911de, line 219: Similar issue as with `N1qlQueryCreator`

mp911de (changes requested):
I left a few comments. Also, you might want to target the `5.0.x` branch with this PR.

mp911de:
Paging @wilkinsona for keeping track of the Couchbase 5.0 progress.

pom.xml
mikereiche: Ok. Removed.

src/main/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreator.java
mikereiche: This was specifically added to handle methods of the type  Long countByxxxx()- the existing code was failing following the the jdk16 enforcement of InaccessibleObjectException.  The code here corresponds to https://github.com/spring-projects/spring-data-mongodb/blob/7a640256698f0772c1509c58e4bc71652a448567/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/repository/query/MongoQueryMethod.java#L143

I'll see if there is a better way to handle this - at least factor out the common code that is repeated in multiple places.

mp911de: yuck! I see, the code in MongoDB isn't even much better although it uses the `EntityMetadata` approach. Creating `PersistentEntity` objects can easily lead to `InaccessibleObjectException` and my comment above is motivated by using code that isn't likely to lead to `InaccessibleObjectException`.

pom.xml
```diff
@@ -14,14 +15,17 @@
...
 
     <properties>
-        <couchbase>3.2.1</couchbase>
-        <couchbase.osgi>3.2.1</couchbase.osgi>
-        <springdata.commons>2.6.0-SNAPSHOT</springdata.commons>
+        <couchbase>3.2.3</couchbase>
+        <couchbase.osgi>3.2.3</couchbase.osgi>
+        <springdata.commons>3.0.0-SNAPSHOT</springdata.commons>
         <java-module-name>spring.data.couchbase</java-module-name>
+        <jodatime>2.10.13</jodatime>
```
mp911de, line 26: We've dropped support for Jodatime in Spring Data Commons. Unless there's a strong reason to keep Joda, we suggest switching to JSR-310 types.

mp911de (approved):
You need to rename `src/main/resources/META-INF/services/javax.enterprise.inject.spi.Extension` to `jakarta.enterprise.inject.spi.Extension`.

Feel free to merge this PR afterward. Thanks for your support!

src/main/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreator.java
mikereiche: The cases where I had made changes were hitting exceptions because of a latent bug - they were using the returnType of the method where they should have been using the entityType. After fixing that bug, the subsequent getPersistentEntity( type ) always returns an existing entity.  I think we are ok here.

pom.xml
mikereiche: We have this as optional - if the user provides the jodatime then we will support it.  I believe the JSR-310 types are handled by https://github.com/spring-projects/spring-data-couchbase/blob/5.0.x/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseJsr310Converters.java

PR #1303, opened 2022-01-15
Allow caching null objects.

### LINKED ISSUE #939: cache.put("key",null) fails [DATACOUCH-628]
**[achoice](https://jira.spring.io/secure/ViewProfile.jspa?name=JIRAUSER50333)** opened **[DATACOUCH-628](https://jira.spring.io/browse/DATACOUCH-628?redirect=false)** and commented

Caching of null values fail with exception thrown by Couchbase client.

CouchbaseCacheConfiguration.defaultCacheConfig() seems that null valus should be cachable. https://github.com/spring-projects/spring-data-couchbase/blob/master/src/main/java/org/springframework/data/couchbase/cache/CouchbaseCacheConfiguration.java#L54

 

```java
@Test
public void cacheNull() {
    final String CACHE_NAME = "CACHE_TEST";
    
    final CouchbaseCacheConfiguration defaultCacheConfig = CouchbaseCacheConfiguration.defaultCacheConfig();
    assertTrue(defaultCacheConfig.getAllowCacheNullValues()); // expect null values to be cachable now

    final CouchbaseCacheManager cacheManager = CouchbaseCacheManager
            .builder(couchbaseClientFactory)
            .cacheDefaults(defaultCacheConfig)
            .initialCacheNames(Set.of(CACHE_NAME))
            .build();
    cacheManager.afterPropertiesSet();

    // TEST cache null
    final Cache cache = cacheManager.getCache(CACHE_NAME);
    cache.put("KEY", null); // exception thrown by couchbase client
}
```

 

---

**Affects:** 4.0.4 (Neumann SR4)

mikereiche:
Use toStoreValue(value) to substitute null objects with NullValue.INSTANCE when putting.  fromStoreValue(value) does the opposite when getting.

Closes #939.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/cache/CacheUser.java
+++ b/src/test/java/org/springframework/data/couchbase/cache/CacheUser.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2022 the original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *        https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.data.couchbase.cache;
+
+import java.io.Serializable;
+
+/**
+ * This is a standalone class (vs. inner) to allow Serialization of all fields to work.
+ * If it was an inner class of CouchbaseCacheIntegrationTests, then it would have a
+ * this$0 field = CouchbaseCacheIntegrationTests and would not serialize.
+ *
+ * @author Michael Reiche
+ */
+class CacheUser implements Serializable {
+	// private static final long serialVersionUID = 8817717605659870262L;
+	String firstname;
+	String lastname;
+	String id;
+
+	public CacheUser(String id, String firstname, String lastname) {
+		this.id = id;
+		this.firstname = firstname;
+		this.lastname = lastname;
+	}
+
+	public String getId() {
+		return id;
+	}
+
+	// define equals for assertEquals()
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/data/couchbase/cache/CacheUser.java
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+package org.springframework.data.couchbase.cache;
+
+import java.io.Serializable;
+
+/**
+ * This is a standalone class (vs. inner) to allow Serialization of all fields to work.
+ * If it was an inner class of CouchbaseCacheIntegrationTests, then it would have a
+ * this$0 field = CouchbaseCacheIntegrationTests and would not serialize.
```
dnault, line 23: If CacheUser were a static nested class, would it it still need to be in a separate file?

https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html

dnault (approved):
LGTM, with some nits.

src/test/java/org/springframework/data/couchbase/cache/CouchbaseCacheIntegrationTests.java
```diff
@@ -0,0 +1,101 @@
...
+
+import com.couchbase.client.java.query.QueryOptions;
+
+/**
+ * CouchbaseCache tests Theses tests rely on a cb server running.
+ *
+ * @author Michael Reiche
+ */
+@IgnoreWhen(clusterTypes = ClusterType.MOCKED)
+class CouchbaseCacheIntegrationTests extends JavaIntegrationTests {
```
dnault, line 40: These tests are good. Are they related to the "Allow caching null objects" change?

Should we have a test that caches and retrieves a null value?

dnault (approved):
LGTM, with some nits.

src/test/java/org/springframework/data/couchbase/cache/CouchbaseCacheIntegrationTests.java
mikereiche: cacheHitMiss() tests a cache hit on a stored null.
It also tests a cache-miss - which is the same for a null as a non-null.
The other tests have cache-hits on non-null (the normal case).

PR #1333, opened 2022-02-10
Add collections support to N1qlJoin.

### LINKED ISSUE #1325: @N1qlJoin and couchbase collections
Does `@N1qlJoin` work with entities annotated with `@Collection`? 
I tried simple join of two entities `@N1qlJoin(on = "lks.rightEntityId=meta(rks).id")` (lks is in different collection than rks) and it does not work. An exception is thrown: `Keyspace not found in CB datastore`. It looks like the default collection is used.

mikereiche:
The scope for the entity can come from an option, a method annotation,
an annotation on the repository or an annotation on the entity class.
All these possibilities are handle by PseudoArgs in the
OperationSupport implementation. That scope/collection are passed
into decodeEntity(). The scope/collection of the child can only come
from an annotation on the entity class. The scope/collection of the
parent and child are uses as follows:

1) Both the parent and the chold have non-default collections
It's possible that the scope for the parent was set with an annotation on
a repository method, the entity class or the repository class or a query
option. Since there is no means to set the scope of the child class by the
method, repository class or query option (only the annotation) we assume
that the (possibly) dynamic scope of the entity would be a better choice
as it is logical to put collections to be joined in the same scope.

2) The parent has a collection (and therefore a scope as well), but the
child  does not have a collection. Use the lhScope and lhCollection for
the entity.  The child is just the bucket.

3) The parent does not have a collection (or scope), but child does have a
collection. Using the same (default) scope for the child would mean
specifying a non-default collection in a default scope - which is not
allowed. So use the scope and collection from the child class.

4) Neither have collections, just use the bucket.

Closes #1325.


src/main/java/org/springframework/data/couchbase/core/ReactiveTemplateSupport.java
```diff
@@ -20,11 +20,15 @@
...
+/**
+ *
+ * @author Michael Reiche
+ */
 public interface ReactiveTemplateSupport {
 
 	Mono<CouchbaseDocument> encodeEntity(Object entityToEncode);
 
-	<T> Mono<T> decodeEntity(String id, String source, long cas, Class<T> entityClass);
+	<T> Mono<T> decodeEntity(String id, String source, long cas, Class<T> entityClass, String scope, String collection);
```
daschl, line 31: I assume our template support classes are not public API so we can break them by extending?

src/main/java/org/springframework/data/couchbase/core/convert/join/N1qlJoinResolver.java
```diff
@@ -111,10 +120,94 @@ public static String buildQuery(ReactiveCouchbaseTemplate template, String colle
...
+		if (lhCollection != null && rhCollection != null) {
+			// they both have non-default collections
+			// It's possible that the scope for the lhs was set with an annotation on a repository method,
+			// the entity class or the repository class or a query option. Since there is no means to set
+			// the scope of the associated class by the method, repository class or query option (only
+			// the annotation) we assume that the (possibly) dynamic scope of the entity would be a better
+			// choice as it is logical to put collections to be joined in the same scope. Note that lhScope
+			// is used for both keyspaces.
+			return new KeySpacePair(lhCollection,
+					x(x("default") + ":" + i(bucketName) + "." + i(lhScope) + "." + i(lhCollection)), //
```
daschl, line 140: is the default: needed? I think n1ql can infer that if not provided? Maybe I'm wrong though..

daschl (commented):
Looks good to me, just a couple comments

src/main/java/org/springframework/data/couchbase/core/ReactiveTemplateSupport.java
mikereiche: Good point. They are public. So I'll keep make a method with the old signature to call this one.

src/main/java/org/springframework/data/couchbase/core/convert/join/N1qlJoinResolver.java
mikereiche: Your are correct. Removed.

PR #1338, opened 2022-02-10
Collection support for predicates on meta fields.

mikereiche:
When a derived query is being created and there is a predicate on
a meta-field ( meta().id, cas, expiry), translate that field
(i.e. id -> meta().id) *without* the bucket or collection name,
as it might apply to a collelction which we do not know yet
(for instance, if there is a withCollection() specified)..
This is fine, because in a derived query, meta() without the
bucket or collection name is unambigous.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreator.java
+++ b/src/main/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreator.java
@@ -76,7 +76,7 @@ public class N1qlQueryCreator extends AbstractQueryCreator<Query, QueryCriteria>
 	protected QueryCriteria create(final Part part, final Iterator<Object> iterator) {
 		PersistentPropertyPath<CouchbasePersistentProperty> path = context.getPersistentPropertyPath(part.getProperty());
 		CouchbasePersistentProperty property = path.getLeafProperty();
-		return from(part, property, where(addMetaIfRequired(bucketName, path, property, entity)), iterator);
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/data/couchbase/repository/query/N1qlQueryCreator.java
```diff
@@ -187,17 +187,28 @@ private QueryCriteria from(final Part part, final CouchbasePersistentProperty pr
...
+	 * @param property
+	 * @param entity
+	 * @return N1QLExpression
+	 */
 	public static N1QLExpression addMetaIfRequired(String bucketName,
 			final PersistentPropertyPath<CouchbasePersistentProperty> persistentPropertyPath,
 			final CouchbasePersistentProperty property, final PersistentEntity entity) {
 		if (entity != null && property == entity.getIdProperty()) {
-			return path(meta(i(bucketName)), i(META_ID_PROPERTY));
+			return path(meta(bucketName != null ? i(bucketName) : x("")), i(META_ID_PROPERTY));
```
daschl, line 205: can the bucket name be empty?

mikereiche: Yes. That's actually what the fix is (line 79 above).  When creating the N1qlQuery (a query derived from a method name), the collection name is not yet available so it just uses meta().id instead of meta(collectionName).id. - which is fine, because in derived queries, meta().id is unambiguous - the collection/bucket is determined by the FROM clause which is constructed later when the collection name is available.

PR #1366, opened 2022-03-11
Support derived queries on repositories defined with an abstract enti…

### LINKED ISSUE #1365: For repositories defined with interface/abstract entity class, only fields in abstract/entity class are projected
For repositories defined with interface/abstract entity class, only fields in abstract/entity class are projected.

To ensure that all the fields for the (unknown at that point) concrete class are projected, then '*' should be projected. 
StringBasedN1qlQueryParser:
```
String projectedFields = i(b) + ".*";
if (resultClass != null  && !Modifier.isAbstract(resultClass.getModifiers()) /* for both interfaces and abstract classes */) {
```		
Also - getFieldName() can be better by returning getName() in the event it cannot access 'field' for the fieldName (which can happen in abstract classes and interfaces).

BaseCouchbasePersistentProperty:
		if(getField() == null){ // punt
			return fieldName = getName();
		}

mikereiche:
…ty class.

Motivation:
Currently an abstract entity class specified in the repository definition can
be used for the predicate typeKey = typeAlias(of abstract entity class) in
queries. Since documents are stored with typeKey = typeAlias(of concrete class)
those queries will never match any documents.  To allow this to work,
all of the abstract entity class an all concrete subclasses must use the
same typeAlias.

Once those documents are found, regardless of their concrete class,
they will all have the same typeKey = typeAlias, instead of having the
typeAlias specific to the concrete class. Additional information in
the stored document is needed to identify the concrete class (subtype
in the example test case), as well as a TypeMapper to interpret that
 information.

Changes:
This allows a common TypeAlias to be used for the purpose of the predicate
typeKey = typeAlias, and the determination of the concrete type by
implementing an AbstractingMappingCouchbaseConverter that inspects the
'subtype' property.

Closes #1365.


mikereiche:
@tchlyah

PR #1403, opened 2022-04-19
Remove explicit dependency on annotations api.

### LINKED ISSUE #1400: Annotation APIs should not have compile scope
Support of QueryDSL has moved `annotation-api` to compile scope, see https://github.com/spring-projects/spring-data-couchbase/commit/c288d053c94cb68121a63395bde7410b1894ece7#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8L163

I don't think it should be a mandatory dependency of this module. Shouldn't it be provided instead? Noticed with `4.4.0-RC1`

mikereiche:
Closes #1400.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -168,14 +168,6 @@
         </dependency>
 
         <dependency>
-            <groupId>javax.annotation</groupId>
-            <artifactId>javax.annotation-api</artifactId>
-            <version>${javax-annotation-api}</version>
-        </dependency>
-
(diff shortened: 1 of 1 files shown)


snicoll:
@mikereiche sorry I can't review this PR. I am merely reporting that a change brought a dependency that wasn't there before. For our own purpose, this commit addresses the problem but I can't say if that's the right fix for this project.

PR #1459, opened 2022-06-15
Add BigDecimal converter.

mikereiche:
Add BigDecimal converter, add converter tests for BigDecimal and BigInteger.
Since BigDecimal is now supported, it cannot be used in the CustomerConverter tests.
So instead use ChoiceFormat for CustomConverter tests.

Closes #1459.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors
+ * Copyright 2021-2022 the original author or authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
```diff
@@ -90,4 +93,24 @@ public BigInteger convert(String source) {
...
 		}
 	}
+
+	@WritingConverter
+	public enum BigDecimalToString implements Converter<BigDecimal, String> {
+		INSTANCE;
+
+		@Override
+		public String convert(BigDecimal source) {
+			return source == null ? null : source.toPlainString();
```
daschl, line 103: Is that how other libraries also do  it instead of using toString()? So serializing it without an exponent field

daschl (commented):
Looks good to me, just one clarification.

src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
mikereiche: That's a really good catch.  That would be incompatible with non- spring data written BigDecimal.

PR #1489, opened 2022-07-08
Use version from parent for all Reactor modules

wilkinsona:
The latest Reactor and Spring Data snapshots are currently incompatible due to some backwards incompatible changes in Reactor. This breaks both Spring Data Couchbase itself and the Couchbase Java client. This has gone unnoticed until now as the Spring Data Couchbase build is not using the version of Reactor that's specified in the Spring Data parent. This pull request corrects this.

Note that the changes in this PR will not build cleanly due to the aforementioned breaking changes.

/cc @simonbasle
/cc @mp911de

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,7 +109,6 @@
         <dependency>
             <groupId>io.projectreactor</groupId>
             <artifactId>reactor-test</artifactId>
-            <version>3.4.6</version>
             <scope>test</scope>
         </dependency>
 


pom.xml
```diff
@@ -39,6 +39,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>io.projectreactor</groupId>
+                <artifactId>reactor-bom</artifactId>
+                <version>${reactor}</version>
```
mp911de, line 45: The import is already present on the Build Parent level, see https://github.com/spring-projects/spring-data-build/blob/5fec771ae2c33ea71fbfc0767ace95d7fbba304b/parent/pom.xml#L845

mp911de (changes requested):

pom.xml
wilkinsona: Interesting. I wonder why the build was still using 3.4.6 then. Is the test-scoped `reactor-test` dependency below with a specific version enough to override the version of `reactor-core` in the compile scope to 3.4.6 too?

wilkinsona: Aah, it wasn't. CI's been broken since the start of the week: https://jenkins.spring.io/view/SpringData/job/spring-data-couchbase/job/main/626/console.

wilkinsona:
> This has gone unnoticed until now as the Spring Data Couchbase build is not using the version of Reactor that's specified in the Spring Data parent. This pull request corrects this.

I was mistaken. It was only the `reactor-test` module that had the wrong version. The build was correctly picking up the incompatible changes as compilation has been [failing since the beginning of the week](https://jenkins.spring.io/view/SpringData/job/spring-data-couchbase/job/main/626/).

PR #1513, opened 2022-07-21
Update transactions documentation for native transaction support.

### LINKED ISSUE #1454: Add Transaction Support to documentation

mikereiche:
Closes #1454.

Staged documentation is available at : https://mikereiche.github.io/staged/transactions.html

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -179,6 +179,29 @@ Building the documentation builds also the project without running tests.
 
 The generated documentation is available from `target/site/reference/html/index.html`.
 
+=== Building and staging reference documentation for review
+
+[source,bash]
+----
+  export MY_GIT_USER=<github-user>
+  mvn generate-resources
+  docs=`pwd`/target/site/reference/html
+  pushd /tmp 
+  mkdir $$
+  cd $$
+  # see https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-github-pages-site
+  # this examples uses a repository named "staged"
+  git clone git@github.com:${MY_GIT_USER}/staged.git -b gh-pages
+  cd staged
+  cp -R $docs/* .
+  git add .
+  git commit --message "stage for review"
+  git push origin gh-pages
+  popd
+----
+
+The generated documentation is available from `target/site/reference/html/index.html`.
+
 == Examples
 
 * https://github.com/spring-projects/spring-data-examples/[Spring Data Examples] contains example projects that explain specific features in more detail.
--- a/src/main/asciidoc/transactions.adoc
+++ b/src/main/asciidoc/transactions.adoc
@@ -1,114 +1,178 @@
 [[couchbase.transactions]]
-= Transaction Support
+= Couchbase Transactions
 
-Couchbase supports https://docs.couchbase.com/server/6.5/learn/data/transactions.html[Distributed Transactions]. This section documents on how to use it with Spring Data Couchbase.
+Couchbase supports https://docs.couchbase.com/server/current/learn/data/transactions.html[Distributed Transactions]. This section documents how to use it with Spring Data Couchbase.
 
 == Requirements
 
- - Couchbase Server 6.5 or above.
- - Couchbase Java client 3.0.0 or above. It is recommended to follow the transitive dependency for the transactions library from maven.
+ - Couchbase Server 6.6.1 or aabove.
+ - Spring Data Couchbase 5.0.0-M5 or above.
  - NTP should be configured so nodes of the Couchbase cluster are in sync with time. The time being out of sync will not cause incorrect behavior, but can impact metadata cleanup.
+ - Set spring.main.allow-bean-definition-overriding=true either in application.properties or as a SpringApplicationBuilder property.
+
+== Overview
+The Spring Data Couchbase template operations insert, find, replace and delete and repository methods that use those calls can participate in a Couchbase Transaction. They can be executed in a transaction by using the @Transactional annotation, the CouchbaseTransactionalOperator, or in the lambda of a Couchbase Transaction.  
 
 == Getting Started & Configuration
 
-The `couchbase-transactions` artifact needs to be included into your `pom.xml` if maven is being used (or equivalent).
+Couchbase Transactions are normally leveraged with a method annotated with @Transactional.
+The @Transactional operator is implemented with the CouchbaseTransactionManager which is supplied as a bean in the AbstractCouchbaseConfiguration.
+Couchbase Transactions can be used without defining a service class by using CouchbaseTransactionOperator which is also supplied as a bean in AbtractCouchbaseConfiguration.
+Couchbase Transactions can also be used directly using Spring Data Couchbase operations within a lambda https://docs.couchbase.com/server/current/learn/data/transactions.html#using-transactions[Using Transactions]
 
- - Group: `com.couchbase.client`
- - Artifact: `couchbase-transactions`
- - Version: latest one, i.e. `1.0.0`
+== Transactions with @Transactional
 
-Once it is included in your project, you need to create a single `Transactions` object. Conveniently, it can be part of
-your spring data couchbase `AbstractCouchbaseConfiguration` implementation:
-
-.Transaction Configuration
+@Transactional defines as transactional a method or all methods on a class. 
+ 
+When this annotation is declared at the class level, it applies as a default
+to all methods of the declaring class and its subclasses.
+ 
+===  Attribute Semantics
+ 
+In this release, the Couchbase Transactions ignores the rollback attributes. 
+The transaction isolation level is read-committed;
+ 
+.Transaction Configuration and Use by @Transactional
 ====
+.The Configuration
 [source,java]
 ----
 @Configuration
+@EnableCouchbaseRepositories("<parent-dir-of-repository-interfaces>")
+@EnableReactiveCouchbaseRepositories("<parent-dir-of-repository-interfaces>")
+@EnableTransactionManagement // <1>
 static class Config extends AbstractCouchbaseConfiguration {
 
-    // Usual Setup
-    @Override public String getConnectionString() { /* ... */ }
-    @Override public String getUserName() { /* ... */ }
-    @Override public String getPassword() { /* ... */ }
-    @Override public String getBucketName() { /* ... */ }
+  // Usual Setup
+  @Override public String getConnectionString() { /* ... */ }
+  @Override public String getUserName() { /* ... */ }
+  @Override public String getPassword() { /* ... */ }
+  @Override public String getBucketName() { /* ... */ }
 
-	@Bean
-	public Transactions transactions(final Cluster couchbaseCluster) {
-		return Transactions.create(couchbaseCluster, TransactionConfigBuilder.create()
-			// The configuration can be altered here, but in most cases the defaults are fine.
-			.build());
-	}
+  // Customization of transaction behavior is via the configureEnvironment() method
+  @Override protected void configureEnvironment(final Builder builder) {
+    builder.transactionsConfig(
+      TransactionsConfig.builder().timeout(Duration.ofSeconds(30)));
+  }
+}
+----
+.The Transactional Service Class
+Note that the body of @Transactional methods can be re-executed if the transaction fails.
+It is imperative that everthing in the method body be idempotent.
+[source,java]
+----
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.Flux;
+
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+final CouchbaseOperations personOperations;
+final ReactiveCouchbaseOperations reactivePersonOperations;
+
+@Service // <2>
+public class PersonService {
+
+  final CouchbaseOperations operations;
+  final ReactiveCouchbaseOperations reactiveOperations;
+
+  public PersonService(CouchbaseOperations ops, ReactiveCouchbaseOperations reactiveOps) {
+    operations = ops;
+    reactiveOperations = reactiveOps;
+  }
+
+  // no annotation results in this method being executed not in a transaction 
+  public Person save(Person p) {
+    return operations.save(p);
+  }
+
+  @Transactional
+  public Person changeFirstName(String id, String newFirstName) {
+    Person p = operations.findById(Person.class).one(id); // <3>
+    return operations.replaceById(Person.class).one(p.withFirstName(newFirstName);
+  }
+
+  @Transactional
+  public Mono<Person> reactiveChangeFirstName(String id, String newFirstName) {
+    return personOperationsRx.findById(Person.class).one(person.id())
+        .flatMap(p -> personOperationsRx.replaceById(Person.class).one(p.withFirstName(newFirstName)));
+  }
 
 }
 ----
+[source,java]
+.Using the @Transactional Service.
+----
+@Autowired PersonService personService; // <4>
+
+Person walterWhite = new Person( "Walter", "White");
+Person p = personService.save(walterWhite); // this is not a transactional method
+...
+Person renamedPerson = personService.changeFirstName(walterWhite.getId(), "Ricky"); // <5>
+----
+Functioning of the  @Transactional method annotation requires 
+[start=1]
+. the configuration class to be annotated with @EnableTransactionManagement;
+. the service object with the annotated methods must be annotated with @Service;
+. the body of the method is executed in a transaction.
+. the service object with the annotated methods must be obtained via @Autowired.
+. the call to the method must be made from a different class than service because calling an annotated
+method from the same class will not invoke the Method Interceptor that does the transaction processing.
 ====
 
-Once the `@Bean` is configured, you can autowire it from your service (or any other class) to make use of it. Please
-see the https://docs.couchbase.com/java-sdk/3.0/howtos/distributed-acid-transactions-from-the-sdk.html[Reference Documentation]
-on how to use the `Transactions` class. Since you need access to the current `Collection` as well, we recommend you to also
-autowire the `CouchbaseClientFactory` and access it from there:
+== Transactions with CouchbaseTransactionalOperator
 
-.Transaction Access
+CouchbaseTransactionalOperator can be used to construct a transaction in-line without creating a service class that uses @Transactional.
+CouchbaseTransactionalOperator is available as a bean and can be instantiated with @Autowired.
+If creating one explicitly, it must be created with CouchbaseTransactionalOperator.create(manager) (NOT TransactionalOperator.create(manager)).
+
+.Transaction Access Using TransactionalOperator.execute()
 ====
 [source,java]
 ----
-@Autowired
-Transactions transactions;
+@Autowired TransactionalOperator txOperator;
+@Autowired ReactiveCouchbaseTemplate reactiveCouchbaseTemplate;
 
-@Autowired
-CouchbaseClientFactory couchbaseClientFactory;
-
-public void doSomething() {
- transactions.run(ctx -> {
-  ctx.insert(couchbaseClientFactory.getDefaultCollection(), "id", "content");
-  ctx.commit();
- });
-}
+Flux<Person> result = txOperator.execute((ctx) -> 
+  reactiveCouchbaseTemplate.findById(Person.class).one(person.id())
+    .flatMap(p -> reactiveCouchbaseTemplate.replaceById(Person.class).one(p.withFirstName("Walt")))
+ );
 ----
 ====
 
-== Object Conversions
+== Transactions Directly with the SDK
 
-Since the transactions library itself has no knowledge of your spring data entity types, you need to convert it back and
-forth when reading/writing to interact properly. Fortunately, all you need to do is autowire the `MappingCouchbaseConverter` and
-utilize it:
+Spring Data Couchbase works seamlessly with the Couchbase Java SDK for transaction processing. Spring Data Couchbase operations that
+can be executed in a transaction will work directly within the lambda of a transactions().run() without involving any of the Spring
+Transactions mechanisms. This is the most straight-forward way to leverage Couchbase Transactions in Spring Data Couchbase.
 
-.Transaction Conversion on Write
+Please see the https://docs.couchbase.com/java-sdk/current/howtos/distributed-acid-transactions-from-the-sdk.html[Reference Documentation]
+
+.Transaction Access - Blocking
 ====
 [source,java]
 ----
-@Autowired
-MappingCouchbaseConverter mappingCouchbaseConverter;
+@Autowired CouchbaseTemplate couchbaseTemplate;
 
-public void doSomething() {
-  transactions.run(ctx -> {
-
-   Airline airline = new Airline("demo-airline", "at");
-   CouchbaseDocument target = new CouchbaseDocument();
-   mappingCouchbaseConverter.write(airline, target);
-
-   ctx.insert(couchbaseClientFactory.getDefaultCollection(), target.getId(), target.getContent());
-
-   ctx.commit();
-  });
-}
+TransactionResult result = couchbaseTemplate.getCouchbaseClientFactory().getCluster().transactions().run(ctx -> {
+  Person p = couchbaseTemplate.findById(Person.class).one(personId);
+  couchbaseTemplate.replaceById(Person.class).one(p.withFirstName("Walt"));
+});
 ----
 ====
 
-The same approach can be used on read:
-
-.Transaction Conversion on Read
+.Transaction Access - Reactive
 ====
 [source,java]
 ----
-TransactionGetResult getResult = ctx.get(couchbaseClientFactory.getDefaultCollection(), "doc-id");
+@Autowired ReactiveCouchbaseTemplate reactiveCouchbaseTemplate;
 
-CouchbaseDocument source = new CouchbaseDocument(getResult.id());
-source.setContent(getResult.contentAsObject());
-Airline read = mappingCouchbaseConverter.read(Airline.class, source);
+Mono<TransactionResult> result = reactiveCouchbaseTemplate.getCouchbaseClientFactory().getCluster().reactive().transactions()
+  .run(ctx -> 
+    reactiveCouchbaseTemplate.findById(Person.class).one(personId)
+      .flatMap(p -> reactiveCouchbaseTemplate.replaceById(Person.class).one(p.withFirstName("Walt")))
+  );
 ----
 ====
 
-We are also looking into tighter integration of the transaction library into the spring data library
-ecosystem.
+
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -51,6 +51,7 @@ import org.springframework.transaction.TransactionManager;
 import org.springframework.transaction.annotation.AnnotationTransactionAttributeSource;
 import org.springframework.transaction.interceptor.TransactionAttributeSource;
 import org.springframework.transaction.interceptor.TransactionInterceptor;
+import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
@@ -330,6 +331,16 @@ public abstract class AbstractCouchbaseConfiguration {
 		return new CouchbaseCallbackTransactionManager(clientFactory);
 	}
 
+	/**
+	 * The default transaction template manager.
+	 *
+	 * @param couchbaseTransactionManager
+	 * @return
+	 */
+	@Bean(BeanNames.COUCHBASE_TRANSACTION_TEMPLATE)
+	TransactionTemplate couchbaseTransactionTemplate(CouchbaseCallbackTransactionManager couchbaseTransactionManager) {
+		return new TransactionTemplate(couchbaseTransactionManager);
+	}
 	/**
 	 * The default TransactionalOperator.
 	 *
--- a/pom.xml
+++ b/pom.xml
@@ -301,8 +301,20 @@
                 <artifactId>maven-assembly-plugin</artifactId>
             </plugin>
             <plugin>
+                <!-- generate asciidoc to stage for review  -->
                 <groupId>org.asciidoctor</groupId>
                 <artifactId>asciidoctor-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>process-asciidoc</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>target/site/reference/html</outputDirectory>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
             <plugin>
                 <groupId>com.mysema.maven</groupId>
--- a/src/main/java/org/springframework/data/couchbase/config/BeanNames.java
+++ b/src/main/java/org/springframework/data/couchbase/config/BeanNames.java
@@ -64,5 +64,7 @@ public class BeanNames {
 
 	public static final String COUCHBASE_TRANSACTION_MANAGER = "couchbaseTransactionManager";
 
+	public static final String COUCHBASE_TRANSACTION_TEMPLATE = "couchbaseTransactionTemplate";
+
 	public static final String COUCHBASE_TRANSACTIONAL_OPERATOR = "couchbaseTransactionalOperator";
 }
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbasePersonTransactionReactiveIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbasePersonTransactionReactiveIntegrationTests.java
@@ -108,7 +108,7 @@ public class CouchbasePersonTransactionReactiveIntegrationTests extends JavaInte
 
 	@Test
 	public void shouldRollbackAfterExceptionOfTxAnnotatedMethod() {
-		assertThrowsWithCause(() -> personService.declarativeSavePersonErrors(WalterWhite).blockLast(),
+		assertThrowsWithCause(() -> personService.declarativeSavePersonErrors(WalterWhite).block(),
 				TransactionSystemUnambiguousException.class, SimulateFailureException.class);
 	}
 
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalNonAllowableOperationsIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalNonAllowableOperationsIntegrationTests.java
@@ -36,7 +36,6 @@ import org.springframework.data.couchbase.util.Capabilities;
 import org.springframework.data.couchbase.util.ClusterType;
 import org.springframework.data.couchbase.util.IgnoreWhen;
 import org.springframework.data.couchbase.util.JavaIntegrationTests;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
@@ -114,9 +113,7 @@ public class CouchbaseTransactionalNonAllowableOperationsIntegrationTests extend
 		});
 	}
 
-	@Service
-	@Component
-	@EnableTransactionManagement
+	@Service // this will work in the unit tests even without @Service because of explicit loading by @SpringJUnitConfig
 	static class PersonService {
 		final CouchbaseOperations personOperations;
 
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalOptionsIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalOptionsIntegrationTests.java
@@ -36,7 +36,6 @@ import org.springframework.data.couchbase.transactions.util.TransactionTestUtil;
 import org.springframework.data.couchbase.util.ClusterType;
 import org.springframework.data.couchbase.util.IgnoreWhen;
 import org.springframework.data.couchbase.util.JavaIntegrationTests;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
@@ -100,9 +99,7 @@ public class CouchbaseTransactionalOptionsIntegrationTests extends JavaIntegrati
 		personService.supportedIsolation();
 	}
 
-	@Service
-	@Component
-	@EnableTransactionManagement
+	@Service // this will work in the unit tests even without @Service because of explicit loading by @SpringJUnitConfig
 	static class PersonService {
 		final CouchbaseOperations ops;
 
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalPropagationIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalPropagationIntegrationTests.java
@@ -43,7 +43,6 @@ import org.springframework.data.couchbase.util.ClusterType;
 import org.springframework.data.couchbase.util.IgnoreWhen;
 import org.springframework.data.couchbase.util.JavaIntegrationTests;
 import org.springframework.lang.Nullable;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 import org.springframework.transaction.IllegalTransactionStateException;
@@ -285,9 +284,7 @@ public class CouchbaseTransactionalPropagationIntegrationTests extends JavaInteg
 		assertEquals(3, attempts.get());
 	}
 
-	@Service
-	@Component
-	@EnableTransactionManagement
+	@Service // this will work in the unit tests even without @Service because of explicit loading by @SpringJUnitConfig
 	static class PersonService {
 		final CouchbaseOperations ops;
 
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalRepositoryIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalRepositoryIntegrationTests.java
@@ -40,7 +40,6 @@ import org.springframework.data.couchbase.transaction.error.TransactionSystemUna
 import org.springframework.data.couchbase.util.ClusterType;
 import org.springframework.data.couchbase.util.IgnoreWhen;
 import org.springframework.data.couchbase.util.JavaIntegrationTests;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
@@ -114,7 +113,6 @@ public class CouchbaseTransactionalRepositoryIntegrationTests extends JavaIntegr
 		String id = UUID.randomUUID().toString();
 
 		assertThrowsWithCause(() -> {
-			;
 			userService.run(repo -> {
 				User user = repo.save(new User(id, "Ada", "Lovelace"));
 				SimulateFailureException.throwEx("fail");
@@ -125,9 +123,7 @@ public class CouchbaseTransactionalRepositoryIntegrationTests extends JavaIntegr
 		assertNull(user);
 	}
 
-	@Service
-	@Component
-	@EnableTransactionManagement
+	@Service // this will work in the unit tests even without @Service because of explicit loading by @SpringJUnitConfig
 	static class UserService {
 		@Autowired UserRepository userRepo;
 
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalTemplateIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalTemplateIntegrationTests.java
@@ -52,7 +52,6 @@ import org.springframework.data.couchbase.util.Capabilities;
 import org.springframework.data.couchbase.util.ClusterType;
 import org.springframework.data.couchbase.util.IgnoreWhen;
 import org.springframework.data.couchbase.util.JavaIntegrationTests;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
@@ -366,9 +365,7 @@ public class CouchbaseTransactionalTemplateIntegrationTests extends JavaIntegrat
 		}, TransactionSystemUnambiguousException.class, IllegalArgumentException.class);
 	}
 
-	@Service
-	@Component
-	@EnableTransactionManagement
+	@Service // this will work in the unit tests even without @Service because of explicit loading by @SpringJUnitConfig
 	static class PersonService {
 		final CouchbaseOperations personOperations;
 		final ReactiveCouchbaseOperations personOperationsRx;
--- a/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalUnsettableParametersIntegrationTests.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/CouchbaseTransactionalUnsettableParametersIntegrationTests.java
@@ -36,10 +36,8 @@ import org.springframework.data.couchbase.util.Capabilities;
 import org.springframework.data.couchbase.util.ClusterType;
 import org.springframework.data.couchbase.util.IgnoreWhen;
 import org.springframework.data.couchbase.util.JavaIntegrationTests;
-import org.springframework.stereotype.Component;
 import org.springframework.stereotype.Service;
 import org.springframework.test.context.junit.jupiter.SpringJUnitConfig;
-import org.springframework.transaction.annotation.EnableTransactionManagement;
 import org.springframework.transaction.annotation.Transactional;
 
 import com.couchbase.client.core.msg.kv.DurabilityLevel;
@@ -202,9 +200,7 @@ public class CouchbaseTransactionalUnsettableParametersIntegrationTests extends 
 		});
 	}
 
-	@Service
-	@Component
-	@EnableTransactionManagement
+	@Service // this will work in the unit tests even without @Service because of explicit loading by @SpringJUnitConfig
 	static class PersonService {
 		final CouchbaseOperations personOperations;
 
--- a/src/test/java/org/springframework/data/couchbase/transactions/PersonService.java
+++ b/src/test/java/org/springframework/data/couchbase/transactions/PersonService.java
@@ -20,7 +20,6 @@ import static com.couchbase.client.java.query.QueryScanConsistency.REQUEST_PLUS;
 import static org.springframework.data.couchbase.util.JavaIntegrationTests.throwSimulateFailureException;
 import static org.springframework.data.couchbase.util.Util.assertInAnnotationTransaction;
 
-import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.util.List;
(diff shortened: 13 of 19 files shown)


src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+.Transaction Access Using TransactionalOperator.execute()
+====
+[source,java]
+----
+@Autowired TransactionalOperator txOperator;
 
-   Airline airline = new Airline("demo-airline", "at");
-   CouchbaseDocument target = new CouchbaseDocument();
-   mappingCouchbaseConverter.write(airline, target);
+    Person person = rxCBTmpl.insertById(Person.class).inCollection(cName).one(WalterWhite).block();
```
daschl, line 219: where is the rxCBTmpl defined? the name is not really good for demonstrative purposes I think

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+.Transaction Access Using transactional()
+====
+[source,java]
+----
+@Autowired TransactionalOperator txOperator;
 
-   ctx.commit();
-  });
-}
+    Person person = rxCBTmpl.insertById(Person.class).inCollection(cName).one(WalterWhite).block();
```
daschl, line 232: same here, is this meant to use txOperator?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
 ====
+.Transaction Access  simulating @Transactional
+====
+[source,java]
+----
+@Service
+@Component
+class PersonService {
+
+    final ReactiveCouchbaseOperations personOperationsRx;
```
daschl, line 247: btw one (maybe subjective) recommendation would be to use Reactive instead of Rx, it could be confused with RxJava and is certainly lingo that some users might not understand

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
 ----
-TransactionGetResult getResult = ctx.get(couchbaseClientFactory.getDefaultCollection(), "doc-id");
+@Autowired CouchbaseTemplate template;
 
-CouchbaseDocument source = new CouchbaseDocument(getResult.id());
-source.setContent(getResult.contentAsObject());
-Airline read = mappingCouchbaseConverter.read(Airline.class, source);
+    TransactionResult result = template.getCouchbaseClientFactory().getCluster().transactions().
+            run(ctx -> 
+                {
```
daschl, line 275: formatting looks a bit off here?

daschl (changes requested):

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+== Transactions with CouchbaseTransactionManager
+
+Note: Reactive Couchbase Transactions cannot be supported via the 
+@Transactional annotation. See the section "Transactions with CouchbaseTransactionalOperator"
+for supporting Reactive Couchbase Transactions using the Transactional Operator.
+ 
+@Transactional Describes a transaction attribute on an individual method or on a class.
+ 
+When this annotation is declared at the class level, it applies as a default
+to all methods of the declaring class and its subclasses. Note that it does not
```
programmatix, line 93: I feel this documentation tends to lead with these lists of caveats and complexities, which I'd find a bit offputting as a user.  Personally I'd lead with simple examples demo-ing the core concepts, and then if necessary explain and demo the caveats and details after.
Also do we need to repeat complexities and rules of Spring @Transactional in our own docs?  It kinda makes our product looks like it's introducing them, not Spring.  I think it's fine to give a basic overview of @Transactional here, but surely the user should be going to the Spring docs for a true understanding and in particular nuances like this?  It would let us simplify our documentation and focus on what we bring to the @transactional table.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+ 
+This annotation is generally directly comparable to Spring's
+org.springframework.transaction.interceptor.RuleBasedTransactionAttribute
+class, and in fact AnnotationTransactionAttributeSource will directly
+convert this annotation's attributes to properties in RuleBasedTransactionAttribute,
+so that Spring's transaction support code does not have to know about annotations.
+ 
+===  Attribute Semantics
+ 
+If no custom rollback rules are configured in this annotation, the transaction
```
programmatix, line 108: I don't think that's the case.  We'll rollback on anything.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+will roll back on RuntimeException and Error but not on checked
+exceptions.
+ 
+Rollback rules determine if a transaction should be rolled back when a given
```
programmatix, line 112: Yeah I don't think we implemented any of this...

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+locally redeclared in order to participate in a subclass-level annotation. For
+details on method visibility constraints, consult the
+https://docs.spring.io/spring-framework/docs/current/reference/html/data-access.html#transaction[Transaction Management]
+section of the reference manual.
+ 
+This annotation is generally directly comparable to Spring's
+org.springframework.transaction.interceptor.RuleBasedTransactionAttribute
+class, and in fact AnnotationTransactionAttributeSource will directly
+convert this annotation's attributes to properties in RuleBasedTransactionAttribute,
+so that Spring's transaction support code does not have to know about annotations.
```
programmatix, line 104: This block is fairly meaningless to me.  I don't understand it and I was directly involved in the implementation :D  What value are we adding to the user with this?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+for the rollback rule. For example, given a rule configured to match against
+"com.example.CustomException", that rule will match against an exception
+named com.example.CustomExceptionV2 (an exception in the same package as
+CustomException but with an additional suffix) or an exception named
+com.example.CustomException$AnotherException (an exception declared as
+a nested class in CustomException).
+ 
+For specific information about the semantics of other attributes in this
+annotation, consult the org.springframework.transaction.TransactionDefinition
+and org.springframework.transaction.interceptor.TransactionAttribute javadocs.
```
programmatix, line 152: Are there permalinks to these available?

programmatix: Do we need to mention TransactionDefinition?  I don't think the user gets access to that in @Transactional?  It's the @transactional attribute documentation that's important here.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+ 
+=== Transaction Management
+ 
+This annotation commonly works with thread-bound transactions managed by a
+org.springframework.transaction.PlatformTransactionManager, exposing a
+transaction to all data access operations within the current execution thread.
+Note: This does NOT propagate to newly started threads within the method.
+
+Couchbase does not supply a reactive transaction manager, therefore it cannot support
+@Transactional on reactive methods. Couchbase provides reactive transaction support
```
programmatix, line 162: But... we do :D  (support it I mean, I know we don't supply a reactive manager)  This is the joy that CouchbaseTransactionInterceptor brings, we wouldn't need it otherwise.
Check out e.g. ReactiveTransactionalTemplateIntegrationTests.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+com.example.CustomException$AnotherException (an exception declared as
+a nested class in CustomException).
+ 
+For specific information about the semantics of other attributes in this
+annotation, consult the org.springframework.transaction.TransactionDefinition
+and org.springframework.transaction.interceptor.TransactionAttribute javadocs.
+ 
+=== Transaction Management
+ 
+This annotation commonly works with thread-bound transactions managed by a
```
programmatix, line 156: What is a thread-bound transaction?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+org.springframework.transaction.PlatformTransactionManager, exposing a
```
programmatix, line 157: As a user how would I know if my transaction is managed by a PlatformTransactionManager?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+transaction to all data access operations within the current execution thread.
```
programmatix, line 158: ... not even sure what this means?

This block is quite a bit of complexity to throw at a user.  Would a simple example not better demo things?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
 ====
 [source,java]
 ----
-@Autowired
-Transactions transactions;
+// this object must be provided by the Spring framework. By @Autowire or from ac.getBean().
+// If created directly with it's constructor, the method interceptor will not be called.
+@Autowired PersonService personService;
+
+    Person p = personService.savePerson(WalterWhite);
```
programmatix, line 173: Whitespace

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+@Component
+public class PersonService {
+
+    final CouchbaseOperations personOperations;
+
+    public PersonService(CouchbaseOperations ops) {
+        personOperations = ops;
+    }
+
+    @Transactional(transactionManager = BeanNames.COUCHBASE_TRANSACTION_MANAGER) // default, shown for clarity
```
programmatix, line 186: Personally I feel the verboseness somewhat decreases clarity here?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+// this object must be provided by the Spring framework. By @Autowire or from ac.getBean().
+// If created directly with it's constructor, the method interceptor will not be called.
+@Autowired PersonService personService;
+
+    Person p = personService.savePerson(WalterWhite);
+
 
-@Autowired
-CouchbaseClientFactory couchbaseClientFactory;
+@Service
```
programmatix, line 176: I think at least the first example we show should be a complete one (though still as small as possible, of course).  E.g. complete classes, and including the config.  Something the user can C&P.  Including a very small entity class if possible.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+@Component
+public class PersonService {
+
+    final CouchbaseOperations personOperations;
+
+    public PersonService(CouchbaseOperations ops) {
```
programmatix, line 182: No @Autowired?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
- transactions.run(ctx -> {
-  ctx.insert(couchbaseClientFactory.getDefaultCollection(), "id", "content");
-  ctx.commit();
- });
 }
 ----
 ====
 
-== Object Conversions
+== Transactions with CouchbaseTransactionalOperator
```
programmatix, line 197: As above, and not to harp on it, but I think we should ditch mention of this, or at least push it to an 'for advanced users' section.  It doesn't add anything to us at all, IMO.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+    public PersonService(CouchbaseOperations ops) {
+        personOperations = ops;
+    }
+
+    @Transactional(transactionManager = BeanNames.COUCHBASE_TRANSACTION_MANAGER) // default, shown for clarity
+    public Person replacePerson(String id) {
+        Person p = personOperations.findById(Person.class).one(id);
+        p.setFirstname(p.getFirstname()+" Junior");
+        return personOperations.replaceById(Person.class).one(p);
+    }
```
programmatix, line 191: Other things we probably want to mention in relation to @Transactional:

- Which operations are supported.
- **Very important**: that this is going to get retried.  They want to fetch their documents inside it, to avoid CAS problems.  They don't want any side-effects (apart from transaction operations of course).
- Do Spring users usually use template methods directly like this, or repository ones?  Maybe want examples for both, leading with the most popular method.

src/main/asciidoc/transactions.adoc
programmatix: Maybe it's here only because you were under the impression it was the only way to do reactive transactions?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
-forth when reading/writing to interact properly. Fortunately, all you need to do is autowire the `MappingCouchbaseConverter` and
-utilize it:
+CouchbaseTransactionalOperator is used for reactive transactions. It is an important class for Couchbase Transactions in Spring Data as there is no Reactive TransactionManager for Couchbase.
+The class simplifies programmatic transaction demarcation and transaction exception handling.
+The central method is transactional(), supporting transactional wrapping of functional sequences code that. This operator handles the transaction lifecycle and possible exceptions such that neither the ReactiveTransactionCallback implementation nor the calling code needs to explicitly handle transactions.
+Typical usage: Allows for writing low-level data access objects that use resources such as database connections but are not transaction-aware themselves. Instead, they can implicitly participate in transactions handled by higher-level application services utilizing this class, making calls to the low-level services via an inner-class callback object.
```
programmatix, line 202: This is a big ol' wall of text that's quite hard to read.  But as above I think we should drop this stuff anyway.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
 
-.Transaction Conversion on Write
+Note that a CouchbaseTransactionalOperator is available as a bean.
+If creating one explicitly, it must be created with the following (and not TransactionalOperator.create(manager)).
+.TransactionalOperator Construction
```
programmatix, line 206: This title gets merged into the previous text.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
 ====
 [source,java]
 ----
-@Autowired
-MappingCouchbaseConverter mappingCouchbaseConverter;
+public static CouchbaseTransactionalOperator create(CouchbaseCallbackTransactionManager manager)
```
programmatix, line 210: NIT: I'd show an example showing how to create it, rather than the method declaration.  Let's never underestimate how much users want to copy, paste and go :)

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+----
+@Autowired TransactionalOperator txOperator;
 
-   ctx.commit();
-  });
-}
+    Person person = rxCBTmpl.insertById(Person.class).inCollection(cName).one(WalterWhite).block();
+    Mono<Person> result = rxCBTmpl.findById(Person.class).one(person.id())
+        .flatMap(p -> rxCBTmpl.replaceById(Person.class).one(p.withFirstName("Walt")))
+        .as(txOperator::transactional);
```
programmatix, line 235: No no no no please let's not have this .as() :)  Hopefully it's not actually in the API and just a docs oversight?  

If not, well...

As we discussed a lot on Slack, I feel this adds a lot of confusion to the user.  It's going to show up to them all the time on Intellisense.  They're going to wonder in what contexts they need to be calling it - inside @Transactional?  Inside a regular SDK transaction?  We're polluting our API for CouchbaseTransactionalOperator which we don't even need..

Plus, it's such a footgun, they can forget it so easily.

And, most importantly, it can't support retries.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+    public Mono<Person> declarativeFindReplacePersonReactive(Person person) {
+        return personOperationsRx.findById(Person.class).one(person.id())
+                .flatMap(p -> personOperationsRx.replaceById(Person.class).one(p.withFirstName(person.getFirstname())))
+                .as(transactionalOperator::transactional);
+    }
+----
+====
 
-.Transaction Conversion on Read
+== Transactions with Only the SDK
```
programmatix, line 266: Oh wow, this sections is very barebones :)

I feel this should be our best seller.  We made the regular transactions API support all Spring operations too - that's really cool!  I think this should be the main way that users interact with transactions in Spring.  We should have this front and centre, and mention @Transactional second - IMO.

Can this section be fleshed out a lot more?  I'd like to see a good description section, and examples including:

- Doing correct error handling of TransactionCommitAmbiguousException and TransactionFailedException (see the docs)
- Mixing the normal transactions API with Spring repo and template methods.

programmatix: Can we not call it "Only the SDK", it just sounds a bit negative :)
Maybe "Using the Transactions SDK Directly" ?  Or something, I dunno, naming things is hard.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
 
-CouchbaseDocument source = new CouchbaseDocument(getResult.id());
-source.setContent(getResult.contentAsObject());
-Airline read = mappingCouchbaseConverter.read(Airline.class, source);
+    TransactionResult result = template.getCouchbaseClientFactory().getCluster().transactions().
+            run(ctx -> 
+                { 
+                    return template.insertById(Person.class).one(WalterWhite); 
+                },
+                options);
```
programmatix, line 278: `options` isn't in this example - in general I think it's best for examples to compile on a C&P

README.adoc
```diff
@@ -179,6 +179,24 @@ Building the documentation builds also the project without running tests.
...
 
+=== Building and staging reference documentation for review
+
+[source,bash]
+----
+  mvn generate-resources
+  docs=`pwd`/target/site/reference/html
+  cd /tmp 
+  # see https://docs.github.com/en/pages/getting-started-with-github-pages/creating-a-github-pages-site
+  git clone git@github.com:mikereiche/staged.git -b gh-pages
```
programmatix, line 190: Probably don't want your name here?

src/main/asciidoc/transactions.adoc
programmatix: We _could_, but I'd probably just document that rollback rules aren't supported, and see if we get any requests for it.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+rollback rules defined via patterns may result in unintentional matches for
+similarly named exceptions and nested classes. This is due to the fact that a
+thrown exception is considered to be a match for a given pattern-based rollback
+rule if the name of thrown exception contains the exception pattern configured
+for the rollback rule. For example, given a rule configured to match against
+"com.example.CustomException", that rule will match against an exception
+named com.example.CustomExceptionV2 (an exception in the same package as
+CustomException but with an additional suffix) or an exception named
+com.example.CustomException$AnotherException (an exception declared as
+a nested class in CustomException).
```
programmatix, line 148: This is a horrendous wall of text, and we're discussing the nitty gritty of rollback rules before we've shown the user a single example of @Transactional in action.  But as above I think we need to replace all this with 'don't work, soz' anyway.

programmatix (changes requested):

src/test/java/org/springframework/data/couchbase/domain/CapellaConnectSample.java
```diff
@@ -90,10 +90,11 @@ public static void main(String... args) {
 		}
 
 		ClusterEnvironment env = ClusterEnvironment.builder()
-				.securityConfig(SecurityConfig.enableTls(true)/*.trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)*/)
+				//.securityConfig(SecurityConfig.enableTls(true)/*.trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)*/)
 				.ioConfig(IoConfig.enableDnsSrv(true)).build();
 
 		// Initialize the Connection
+		endpoint = "cb.zsibzkbgllfbcj8g.cloud.couchbase.com";
```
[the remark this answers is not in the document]
programmatix: And the example should include couchbases:// so TLS gets enabled and it just works.

src/test/java/org/springframework/data/couchbase/domain/CapellaConnectSample.java
```diff
@@ -90,10 +90,11 @@ public static void main(String... args) {
 		}
 
 		ClusterEnvironment env = ClusterEnvironment.builder()
-				.securityConfig(SecurityConfig.enableTls(true)/*.trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)*/)
+				//.securityConfig(SecurityConfig.enableTls(true)/*.trustManagerFactory(InsecureTrustManagerFactory.INSTANCE)*/)
```
programmatix, line 93: Do we want this commenting left in the example?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+will roll back on RuntimeException and Error but not on checked
+exceptions.
+ 
+Rollback rules determine if a transaction should be rolled back when a given
+exception is thrown, and the rules are based on types or patterns. Custom
+rules may be configured via rollbackFor/noRollbackFor and
+#rollbackForClassName/noRollbackForClassName, which allow
+rules to be specified as types or patterns, respectively.
+ 
+When a rollback rule is defined with an exception type, that type will be
```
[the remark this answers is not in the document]
programmatix: nope :)

src/main/asciidoc/transactions.adoc
mikereiche: It's the reactiveCouchbaseTemplate.  I'll give it a better name and show the definition (@Autowired).

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -330,6 +331,16 @@ CouchbaseCallbackTransactionManager couchbaseTransactionManager(CouchbaseClientF
 		return new CouchbaseCallbackTransactionManager(clientFactory);
 	}
 
+	/**
+	 * The default transaction template manager.
+	 *
+	 * @param couchbaseTransactionManager
+	 * @return
+	 */
+	@Bean(BeanNames.COUCHBASE_TRANSACTION_TEMPLATE)
```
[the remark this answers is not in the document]
mikereiche: I missed defining it in the original Transactions implementation branch.  I only noticed it was missing when I was writing the doc samples.

README.adoc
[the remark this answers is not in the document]
mikereiche: No. It needs to reference a user-defined repository, created by following the instructions in the link above.  I should make that clear.
The github-pages for spring-data-couchbase is already used - it redirects to the actual spring-data-couchbase documentation.

src/test/java/org/springframework/data/couchbase/domain/CapellaConnectSample.java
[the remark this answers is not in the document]
mikereiche: Oops.  This file shouldn't be in the branch. Someone ask a Capella question and I used my current branch to try it out. And then 'git added' everything. 

> And the example should include couchbases:// so TLS gets enabled and it just works.

That's a good point.  I will put that in a separate branch for the Capella example.

src/main/asciidoc/transactions.adoc
mikereiche: That's a really good point.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,30 +1,38 @@
 [[couchbase.transactions]]
+Couchbase Transactions
 = Transaction Support
 
-Couchbase supports https://docs.couchbase.com/server/6.5/learn/data/transactions.html[Distributed Transactions]. This section documents on how to use it with Spring Data Couchbase.
+Couchbase supports https://docs.couchbase.com/server/6.5/learn/data/transactions.html[Distributed Transactions]. This section documents how to use it with Spring Data Couchbase.
```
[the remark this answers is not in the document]
mikereiche: Good point. I was thinking it should be the earliest version where it works, but it should indeed be "current".

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
-on how to use the `Transactions` class. Since you need access to the current `Collection` as well, we recommend you to also
-autowire the `CouchbaseClientFactory` and access it from there:
+Please see the https://docs.couchbase.com/java-sdk/current/howtos/distributed-acid-transactions-from-the-sdk.html[Reference Documentation]
+
+== Transactions with TransactionTemplate
+The TransactionTemplate class simplifies programmatic transaction demarcation and transaction exception handling.
+The central method is execute, supporting transactional code that implements the TransactionCallback interface. This template handles the transaction lifecycle and possible exceptions such that neither the TransactionCallback implementation nor the calling code needs to explicitly handle transactions.
+Typical usage: Allows for writing low-level data access objects that use resources such as JDBC DataSources but are not transaction-aware themselves. Instead, they can implicitly participate in transactions handled by higher-level application services utilizing this class, making calls to the low-level services via an inner-class callback object.
```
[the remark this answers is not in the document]
mikereiche: Agree.  A little too much "cut-and-paste" from generic documentation.  I'll rework it.

src/main/asciidoc/transactions.adoc
mikereiche: reactiveCouchbaseTemplate

src/main/asciidoc/transactions.adoc
mikereiche: Yes, I tried to call out the block {...} of operations in the transaction, but it just ends up being ugly.  I'll let the code formatter handle it.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,30 +1,38 @@
 [[couchbase.transactions]]
+Couchbase Transactions
 = Transaction Support
 
-Couchbase supports https://docs.couchbase.com/server/6.5/learn/data/transactions.html[Distributed Transactions]. This section documents on how to use it with Spring Data Couchbase.
+Couchbase supports https://docs.couchbase.com/server/6.5/learn/data/transactions.html[Distributed Transactions]. This section documents how to use it with Spring Data Couchbase.
 
 == Requirements
 
  - Couchbase Server 6.5 or above.
```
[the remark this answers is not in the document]
mikereiche: Right. Same as above - while the "6.5 or above" is correct for "Requirements", this should always point to the current documentation.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+----
+====
+
+== Transactions with CouchbaseTransactionManager
+
+Note: Reactive Couchbase Transactions cannot be supported via the 
+@Transactional annotation. See the section "Transactions with CouchbaseTransactionalOperator"
+for supporting Reactive Couchbase Transactions using the Transactional Operator.
+ 
+@Transactional Describes a transaction attribute on an individual method or on a class.
```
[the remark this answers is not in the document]
mikereiche: It's a poor choice of words (especially since there is a TransactionAttribute object!). Better would be "@Transactional defines the method (or methods of the class) as executing in a Couchbase Transaction (by means of a method interceptor that leverages a Couchbase Transaction Manager).

src/main/asciidoc/transactions.adoc
mikereiche: Not with our current transaction manager. Other implementations have a TransactionDefinition argument to their transaction manager constructor, but we don't. (We used to, but removed it as it served no purpose).

src/test/java/org/springframework/data/couchbase/domain/CapellaConnectSample.java
mikereiche: Oops.  This file shouldn't be in this branch.

src/main/asciidoc/transactions.adoc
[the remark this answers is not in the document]
programmatix: No I mean the minimum version is 6.6.1 now :)

src/main/asciidoc/transactions.adoc
mikereiche: Oh. I kind of missed that.  So TransactionalOperator is not necessary for reactive transactions.  I can relegate TransactionalOperator to an "Advanced" section (or just remove it).

src/main/asciidoc/transactions.adoc
mikereiche: removed.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
+. the configuration class to be annotated with @EnableTransactionManagement;
+. the service object with the annotated methods must be annotated with @Service and @Component;
+. the service object with the annotated methods must be obtained as an @Bean either via @Autowire or
+from the application context;
+. the call to the method must be made from a different class than service because calling an annotated
+method from the same class will not invoke the Method Interceptor that does the transaction processing.
+
+== Transactions with @Transactional
+
+@Transactional Defines as transactional a method or all methods on a class.
```
programmatix, line 56: Still wants to be lower case D I think :)

src/main/asciidoc/transactions.adoc
programmatix: I still feel this following the changes :)

Do we have to lead with details of exactly how @Transactional applies to a class hierarchy?  It's straight into the weeds and into yawnsville for the reader... If we have to mention them, could we push them to later?  Or preferable, just figure that we don't need to re-explain how Spring's @Transactional work, and just focus on what we bring to the table?

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
+details on method visibility constraints, consult the
+https://docs.spring.io/spring-framework/docs/current/reference/html/data-access.html#transaction[Transaction Management]
+section of the reference manual.
+ 
+===  Attribute Semantics
+ 
+In this release, he Couchbase Transactions implementation in 
+Spring Data Couchbase ignores all @Transactional attributes.
+The transaction will roll back on uncaught exceptions. 
+The transaction isoloation level is read committed;
```
programmatix, line 71: Lot of typos here.

And "In this release, he Couchbase Transactions implementation in Spring Data Couchbase ignores all @Transactional attributes." - it doesn't.  I made it support most of them - timeout, propagation levels, etc.  I think it's only rollback rules we don't.

programmatix: Well we reject at runtime non-supported isolation levels, but I think that's allowed.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
+Non-blocking transactions are bound to the reactive context.
 
 .Transaction Access
 ====
 [source,java]
 ----
-@Autowired
-Transactions transactions;
+// this object must be provided by the Spring framework either by @Autowire or from ac.getBean().
+// If created directly with its constructor, the method interceptor will not be called.
```
programmatix, line 83: What is `ac` here?
Do we want to keep going into the weeds with why things don't work?
Also I think this is perhaps a standard Spring limitation, that constructing it directly doesn't create a proxy and so @Transactional probably doesn't work (I'm not 100% certain).  Is it something we want to mention in our own docs?

src/main/asciidoc/transactions.adoc
programmatix: I still feel this following the changes :)

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
+
+@Service
+@Component
+public class PersonService {
+
+  final CouchbaseOperations operations;
+  final ReactiveCouchbaseOperations reactiveOperations;
+
+  // This method will be called for @Autowired using the 
+  // CouchbaseOperations and ReactiveCouchbaseOperations @Beans
```
programmatix, line 109: I don't really understand what this sentence means?  Admittedly, am a Spring noob.
Does it not need @Autowired on the constructor for DI to work?

src/main/asciidoc/transactions.adoc
programmatix: Still does

src/main/asciidoc/transactions.adoc
programmatix: Same feedback

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
-MappingCouchbaseConverter mappingCouchbaseConverter;
+@Autowired TransactionalOperator txOperator;
+@Autowired ReactiveCouchbaseTemplate reactiveCouchbaseTemplate;
 
-public void doSomething() {
-  transactions.run(ctx -> {
+Person person = reactiveCouchbaseTemplate.insertById(Person.class).inCollection(cName).one(WalterWhite).block();
+Flux<Person> result = txOperator.execute((ctx) -> reactiveCouchbaseTemplate.findById(Person.class).one(person.id())
+  .flatMap(p -> reactiveCouchbaseTemplate.replaceById(Person.class).one(p.withFirstName("Walt"))));
+result.blockLast();
```
programmatix, line 159: Example formatting is a bit messy and hard to read

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
+----
+====
 
-   Airline airline = new Airline("demo-airline", "at");
-   CouchbaseDocument target = new CouchbaseDocument();
-   mappingCouchbaseConverter.write(airline, target);
+== Transactions Directly with the SDK
```
programmatix, line 163: So as in previous review

1. I think this should be front and centre as the Couchbase recommended way to do transactions with Spring, and mentioned right at the top.  Full power, no limitations.
2. It's too short :)  We don't want to C&P the docs but we can mention what's cool about this.  We can say that you can use any Spring operations directly inside the lambda, that you can mix and match those with the regular transactional API, that sort of thing.  Let's big ourselves up a bit :)

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
+    .run(ctx -> TransactionalSupport.checkForTransactionInThreadLocalStorage().then(Mono.defer(() -> {
+      reactiveCouchbaseTemplate.insertById(Person.class).one(WalterWhite);
+    }))).block();
 ----
 ====
 
-We are also looking into tighter integration of the transaction library into the spring data library
-ecosystem.
\ No newline at end of file
+
```
programmatix, line 191: Is this the end of the docs now?
There's an awful lot of feedback from before that hasn't been incorporated... including very very crucial things like talking about the retry model.  Could you please take another pass on my previous review?  Anything you disagree with and don't want to incorporate, please comment on it and we can iterate further.

programmatix (changes requested):

src/main/asciidoc/transactions.adoc
mikereiche: btw - this would be a good place to use callouts in the sample to point to this constructor and above where the @Autowired appears. `@Autowired PersonService personService`

src/main/asciidoc/transactions.adoc
mikereiche: It's what the spring-data specified formatter gives. I tried formatting these to show the different parts of the transactions (find followed by replace in this case), but Michael complained.

src/main/asciidoc/transactions.adoc
mikereiche: The link to the transactions doc is there. 
> We can say that you can use any Spring operations directly inside the lambda

Yes - I seem to have copied only the samples and lost the accompanying text. Will revisit that.

src/main/asciidoc/transactions.adoc
mikereiche: I'll look at the review comments again.
I'll try to describe the retry model.  But on the one hand there are a bunch of comments about not getting bogged down in the weeds and implementation details, and on the other hand "there's no explanation for xyz".

src/main/asciidoc/transactions.adoc
```diff
@@ -1,30 +1,38 @@
...
- - Group: `com.couchbase.client`
- - Artifact: `couchbase-transactions`
- - Version: latest one, i.e. `1.0.0`
+The Couchbase Transaction Manager, Couchbase Transactional Operator and Couchbase Transaction Template(5.0.0-M6+)  are available as beans from the Configuration class that 
+extends AbstractCouchbaseConfiguration.
 
-Once it is included in your project, you need to create a single `Transactions` object. Conveniently, it can be part of
-your spring data couchbase `AbstractCouchbaseConfiguration` implementation:
+Functioning of the  @Transactional method annotation requires the configuration class to be annotated 
+with @EnableTransactionManagement. The service object with the annotated methods must be obtained as an @Bean either via @Autowire or
```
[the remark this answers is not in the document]
programmatix: Yep this is much easier to read now, especially with the cross-referenced numbers - thanks.

src/main/asciidoc/transactions.adoc
programmatix: It's more the "lead with" part that was offputting.  The docs are much better on this now after a couple of revisions, but they previously included huge chunks of texts regarding specifics of niche @Transactional attributes, and that kind of thing - all long before we got to just showing how to use it.

I totally accept that sometimes we need to mention some caveats and things for the user to be aware of - like I initially mentioned, "Personally I'd lead with simple examples demo-ing the core concepts, and then if necessary explain and demo the caveats and details after."

I'll comment on the @service thing in that part of the review, but I wasn't looking for an explanation of it as such.

src/main/asciidoc/transactions.adoc
programmatix: The current text is "In this release, the Couchbase Transactions ignores the rollback attributes. The transaction isolation level is read-committed;"

Looks like should be a full stop not a semicolon at end, and the first sentence doesn't look right.  Maybe "The @Transactional `rollback` parameter is ignored."?

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+The central method is execute, supporting transactional code that implements the TransactionCallback interface. This template handles the transaction lifecycle and possible exceptions such that neither the TransactionCallback implementation nor the calling code needs to explicitly handle transactions.
+Typical usage: Allows for writing low-level data access objects that use resources such as JDBC DataSources but are not transaction-aware themselves. Instead, they can implicitly participate in transactions handled by higher-level application services utilizing this class, making calls to the low-level services via an inner-class callback object.
+Can be used within a service implementation via direct instantiation with a transaction manager reference, or get prepared in an application context and passed to services as bean reference. Note: The transaction manager should always be configured as bean in the application context: in the first case given to the service directly, in the second case given to the prepared template.
+Supports setting the propagation behavior and the isolation level by name, for convenient configuration in context definitions.
```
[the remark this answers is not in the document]
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+        Person p = personOperations.findById(Person.class).one(id);
+        p.setFirstname(p.getFirstname()+" Junior");
+        personOperations.replaceById(Person.class).one(p);
+    });
+----
+====
+
+== Transactions with CouchbaseTransactionManager
+
+Note: Reactive Couchbase Transactions cannot be supported via the
```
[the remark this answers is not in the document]
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
```diff
@@ -33,82 +41,257 @@ static class Config extends AbstractCouchbaseConfiguration {
...
+
+    transactionTemplate.executeWithoutResult(status -> {
+        Person p = personOperations.findById(Person.class).one(id);
+        p.setFirstname(p.getFirstname()+" Junior");
+        personOperations.replaceById(Person.class).one(p);
+    });
+----
+====
+
+== Transactions with CouchbaseTransactionManager
```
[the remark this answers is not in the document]
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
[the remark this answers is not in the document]
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now resolved.

src/main/asciidoc/transactions.adoc
programmatix: Heh.  A fair point. 

But, Spring is significantly more complicated than the SDK.  There's dependency injection, configuration classes ... - it's a lot less works as you'd expect out of the box IMO.

FWIW I think the revised example is pretty nice.  I can easily follow it now, the number cross references are good, it demos all the crucial concepts concisely.

src/main/asciidoc/transactions.adoc
programmatix: I'll just assume this does work then and the dependency injection happens even without Autowired.

src/main/asciidoc/transactions.adoc
programmatix: The middle one is now done, though it visually seems a bit buried to me in the midst of the initial example.

The other two aren't AFAICT, but they are less critical.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, this is now done.

src/main/asciidoc/transactions.adoc
programmatix: Yep and the 2nd example includes "@Autowired TransactionalOperator txOperator;" which is great.  The inclusion of this "TransactionalOperator Construction" section would leave me confused - it's implying that I need to create it, but then the text and the example show it autowired.

But this was raised as a NIT anyway, I'll let it go.

src/main/asciidoc/transactions.adoc
programmatix: For my reference, mention of .as() is now resolved.
And hopefully isn't in the API itself...

src/main/asciidoc/transactions.adoc
programmatix: I still feel like we could talk this section up a bit, it's incredibly barebones.  But I guess we just disagree on this one.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
 
-The `couchbase-transactions` artifact needs to be included into your `pom.xml` if maven is being used (or equivalent).
-
- - Group: `com.couchbase.client`
- - Artifact: `couchbase-transactions`
- - Version: latest one, i.e. `1.0.0`
-
-Once it is included in your project, you need to create a single `Transactions` object. Conveniently, it can be part of
-your spring data couchbase `AbstractCouchbaseConfiguration` implementation:
+Couchbase Transactions are normally leveraged with the @Transactional operator.
```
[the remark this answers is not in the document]
programmatix: Ok.  We disagree on this, but it's ok, we just have different priorities.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
-
-	@Bean
-	public Transactions transactions(final Cluster couchbaseCluster) {
-		return Transactions.create(couchbaseCluster, TransactionConfigBuilder.create()
-			// The configuration can be altered here, but in most cases the defaults are fine.
-			.build());
-	}
-
+// Usual Setup
+@Override public String getConnectionString() { /* ... */ }
```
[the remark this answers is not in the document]
programmatix: Can't fully recall but I think it was at the start of the line, rather than indented as you'd expect inside a class.  Looks all good now anyway.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
 
-Once the `@Bean` is configured, you can autowire it from your service (or any other class) to make use of it. Please
-see the https://docs.couchbase.com/java-sdk/3.0/howtos/distributed-acid-transactions-from-the-sdk.html[Reference Documentation]
-on how to use the `Transactions` class. Since you need access to the current `Collection` as well, we recommend you to also
-autowire the `CouchbaseClientFactory` and access it from there:
+
+Functioning of the  @Transactional method annotation requires 
+[start=1]
+. the configuration class to be annotated with @EnableTransactionManagement;
+. the service object with the annotated methods must be annotated with @Service and @Component;
```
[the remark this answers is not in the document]
programmatix: Yeah I wasn't asking for a tutorial on what a service is, or anything extremely verbose like the proxy object thing.  Just that the text made it clearer you were referring to a @Service.  The new numbering cross-referenced to the code is fine.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,191 @@
...
 ====
 
-Once the `@Bean` is configured, you can autowire it from your service (or any other class) to make use of it. Please
-see the https://docs.couchbase.com/java-sdk/3.0/howtos/distributed-acid-transactions-from-the-sdk.html[Reference Documentation]
-on how to use the `Transactions` class. Since you need access to the current `Collection` as well, we recommend you to also
-autowire the `CouchbaseClientFactory` and access it from there:
+
+Functioning of the  @Transactional method annotation requires 
+[start=1]
+. the configuration class to be annotated with @EnableTransactionManagement;
```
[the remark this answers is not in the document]
programmatix: I'm not sure if it's asciidoctor or antora that understands this syntax, but I use "// <1>" in the transactions docs here https://docs.couchbase.com/java-sdk/current/howtos/distributed-acid-transactions-from-the-sdk.html#read-your-own-writes, and it renders a nice number bubble.

src/main/asciidoc/transactions.adoc
programmatix: Heh :)

src/main/asciidoc/transactions.adoc
programmatix: Ok.

src/main/asciidoc/transactions.adoc
programmatix: Sure, I get all that (the DI stuff).  Github reviewing/commenting is getting kind of confusing and it's not letting me see enough of the original snippet I commented on to be sure of the context here.  Let's assume it's resolved.

src/main/asciidoc/transactions.adoc
programmatix: That's a shame, it's still quite hard to read.

src/main/asciidoc/transactions.adoc
programmatix: I mean, yes - there were niche concerns that were extremely detailed or didn't apply to us at all (it's much better now), and crucial aspects that weren't described.

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,179 @@
...
 ----
-TransactionGetResult getResult = ctx.get(couchbaseClientFactory.getDefaultCollection(), "doc-id");
+@Autowired ReactiveCouchbaseTemplate reactiveCouchbaseTemplate;
 
-CouchbaseDocument source = new CouchbaseDocument(getResult.id());
-source.setContent(getResult.contentAsObject());
-Airline read = mappingCouchbaseConverter.read(Airline.class, source);
+TransactionResult result = reactiveCouchbaseTemplate.getCouchbaseClientFactory().getCluster().reactive().transactions()
+    .run(ctx -> TransactionalSupport.checkForTransactionInThreadLocalStorage().then(Mono.defer(() -> {
+      reactiveCouchbaseTemplate.insertById(Person.class).one(WalterWhite);
```
programmatix, line 174: Shouldn't include checkForTransactionInThreadLocalStorage right?

src/main/asciidoc/transactions.adoc
```diff
@@ -1,114 +1,179 @@
...
+method from the same class will not invoke the Method Interceptor that does the transaction processing.
 ====
 
-Once the `@Bean` is configured, you can autowire it from your service (or any other class) to make use of it. Please
-see the https://docs.couchbase.com/java-sdk/3.0/howtos/distributed-acid-transactions-from-the-sdk.html[Reference Documentation]
-on how to use the `Transactions` class. Since you need access to the current `Collection` as well, we recommend you to also
-autowire the `CouchbaseClientFactory` and access it from there:
+== Transactions with CouchbaseTransactionalOperator
+
+CouchbaseTransactionalOperator can be used to construct a tansaction in-line without creating a service class that uses @Transactional.
```
programmatix, line 125: typo

programmatix (changes requested):

src/main/asciidoc/transactions.adoc
mikereiche: I looked at what Maria did with the main transactions documentation.   I want her to give this a going-over when she gets a chance. (also the whole spring-data-couchbase documentation - it's all very ... minimal).

src/main/asciidoc/transactions.adoc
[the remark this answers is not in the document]
mikereiche: Oh - I think I know.  I copied right from the java file and removed the lines I didn't want - which resulted in the imports having no indenting and the code having indenting (since it was in a method) which looks weird in a sample.

src/main/asciidoc/transactions.adoc
programmatix: Yeah that's a great idea, Maria will likely have some good suggestions.

src/main/asciidoc/transactions.adoc
mikereiche: removed.

src/main/asciidoc/transactions.adoc
mikereiche: The magic happens with

@Autowired PersonService personService;

spring looks for candidates for the Autowiring.  It finds that there is an @Service PersonService.  It examines the constructor for PersonService and sees that it takes a CouchbaseOperation arg.  It looks for candidates for that, and finds that there is an @Bean CouchbaseTemplate and that CouchbaseTemplate implements CouchbaseOperations - so it uses that.

src/main/asciidoc/transactions.adoc
mikereiche: Aw crap. I had mentioned how "Warning:" made a nice box, but then I just stuck it in the code example.

src/main/asciidoc/transactions.adoc
mikereiche: I stuck in some extra text to brighten it up.

src/main/asciidoc/transactions.adoc
programmatix: Ah interesting, I didn't know the DI could propagate through like that, e.g. is only required on the initial field/ctor - I thought each constructor (or field) invoked also needed the @Autowired to work.

src/main/asciidoc/transactions.adoc
programmatix: Heh

PR #1514, opened 2022-07-21
Reorganize the github repository.

### LINKED ISSUE #1503: reorganize repository into code and sample modules

mikereiche:
Reorganize the github repository to allow addition of a sample/test application
that consumes spring-data-couchbase.

Closes #1503.


mikereiche:
Hi @mp911de  - The main purpose is to include a spring-boot application that consumes spring-data-couchbase. We had an issue last year where I merged a contributed PR that removed a public api that broke the application initialization and we didn't catch that because we only have tests that have application initialization via @SpringJUnitConfig(Config.class) instead of a real spring-boot application.

mp911de:
Having a dependency to Spring Boot does not really work because of the dependency graph. Conceptually, Boot depends on Data and introducing a Boot dependency into a Data repository breaks the separation. If you want to verify the integration in an automated way, I suggest setting up a separate repository that consumes the latest data snapshots.

mikereiche:
Ugh.  That's kind of important. Thanks.

mikereiche:
> Having a dependency to Spring Boot does not really work because of the dependency graph

The intent is to not have a &lt;module&gt; in the parent pom.xml for the sample application - and therefore no cyclic dependency.  It would just be located in the same repository, so when cloning spring-data-couchbase, the sample app would be available.   It's true that the same could be accomplished in a separate repository, but this would be more convenient and more likely to be used by someone working on spring-data-couchbase.

PR #1528, opened 2022-08-05
Fix call to block() in CouchbaseCallbackTransactionManager.

### LINKED ISSUE #1527: reactive @Transactional fails with java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking
From #1524.

The underlying issue is that TransactionInterceptor is being called instead of CouchbaseTransactionInterceptor. And it should be calling executeReactive() instead of execute().

```
@Service
class ScheduleService(
    private val reactiveOperations: ReactiveCouchbaseOperations
) {
    @Transactional
    fun saveScheduleWithTransaction(schedule: Schedule): Mono<Schedule> {
        return reactiveOperations.insertById(Schedule::class.java).one(schedule)
    }
}
```

```
java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread reactor-http-nio-3
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:83) ~[reactor-core-3.5.0-M4.jar:3.5.0-M4]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	*__checkpoint ⇢ org.springframework.boot.actuate.metrics.web.reactive.server.MetricsWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ HTTP POST "/schedule" [ExceptionHandlingWebHandler]
Original Stack Trace:
		at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:83) ~[reactor-core-3.5.0-M4.jar:3.5.0-M4]
		at reactor.core.publisher.Mono.block(Mono.java:1675) ~[reactor-core-3.5.0-M4.jar:3.5.0-M4]
		at org.springframework.data.couchbase.transaction.CouchbaseCallbackTransactionManager.handlePropagation(CouchbaseCallbackTransactionManager.java:191) ~[spring-data-couchbase-5.0.0-M5.jar:5.0.0-M5]
		at org.springframework.data.couchbase.transaction.CouchbaseCallbackTransactionManager.execute(CouchbaseCallbackTransactionManager.java:76) ~[spring-data-couchbase-5.0.0-M5.jar:5.0.0-M5]
		at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:417) ~[spring-tx-6.0.0-M5.jar:6.0.0-M5]
		at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:119) ~[spring-tx-6.0.0-M5.jar:6.0.0-M5]
		at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:184) ~[spring-aop-6.0.0-M5.jar:6.0.0-M5]
		at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:763) ~[spring-aop-6.0.0-M5.jar:6.0.0-M5]
		at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:708) ~[spring-aop-6.0.0-M5.jar:6.0.0-M5]
		at com.rbleuse.spring.reactive.couchbase.service.ScheduleService$$EnhancerBySpringCGLIB$$f091fb23.saveScheduleWithTransaction(<generated>) ~[main/:na]
		at com.rbleuse.spring.reactive.couchbase.handler.ScheduleHandler.createSchedule(ScheduleHandler.kt:20) ~[main/:na]
```

mikereiche:
Closes #1527.

### DIFF
--- a/spring-data-couchbase/src/main/java/org/springframework/data/couchbase/core/TransactionalSupport.java
+++ b/spring-data-couchbase/src/main/java/org/springframework/data/couchbase/core/TransactionalSupport.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.data.couchbase.core;
 
+import com.couchbase.client.core.transaction.threadlocal.TransactionMarker;
 import reactor.core.publisher.Mono;
 
 import java.util.Optional;
@@ -26,6 +27,7 @@ import com.couchbase.client.core.error.CasMismatchException;
 import com.couchbase.client.core.error.transaction.TransactionOperationFailedException;
 import com.couchbase.client.core.transaction.CoreTransactionAttemptContext;
 import com.couchbase.client.core.transaction.threadlocal.TransactionMarkerOwner;
+import reactor.util.context.ContextView;
 
 /**
  * Utility methods to support transactions.
@@ -49,6 +51,14 @@ public class TransactionalSupport {
 			return Mono.just(out);
 		});
 	}
+
+	public static Optional<CouchbaseResourceHolder> checkForTransactionInThreadLocalStorage(ContextView ctx) {
+		return Optional.ofNullable(ctx.hasKey(TransactionMarker.class) ? new CouchbaseResourceHolder(ctx.get(TransactionMarker.class).context()) : null);
+	}
+
+	//public static Optional<CouchbaseResourceHolder> blockingCheckForTransactionInThreadLocalStorage() {
+	//		return TransactionMarkerOwner.marker;
+	//	}
 
 	public static Mono<Void> verifyNotInTransaction(String methodName) {
 		return checkForTransactionInThreadLocalStorage().flatMap(s -> {
--- a/spring-data-couchbase/src/main/java/org/springframework/data/couchbase/transaction/CouchbaseCallbackTransactionManager.java
+++ b/spring-data-couchbase/src/main/java/org/springframework/data/couchbase/transaction/CouchbaseCallbackTransactionManager.java
@@ -44,6 +44,7 @@ import com.couchbase.client.java.transactions.TransactionResult;
 import com.couchbase.client.java.transactions.config.TransactionOptions;
 import com.couchbase.client.java.transactions.error.TransactionCommitAmbiguousException;
 import com.couchbase.client.java.transactions.error.TransactionFailedException;
+import reactor.util.context.ContextView;
 
 /**
  * The Couchbase transaction manager, providing support for @Transactional methods.
@@ -73,7 +74,7 @@ public class CouchbaseCallbackTransactionManager implements CallbackPreferringPl
 
 	@Override
 	public <T> T execute(TransactionDefinition definition, TransactionCallback<T> callback) throws TransactionException {
-		boolean createNewTransaction = handlePropagation(definition);
+		boolean createNewTransaction = handlePropagation(definition, null);
 
 		setOptionsFromDefinition(definition);
 
@@ -87,8 +88,8 @@ public class CouchbaseCallbackTransactionManager implements CallbackPreferringPl
 	@Stability.Internal
 	<T> Flux<T> executeReactive(TransactionDefinition definition,
 			org.springframework.transaction.reactive.TransactionCallback<T> callback) {
-		return Flux.defer(() -> {
-			boolean createNewTransaction = handlePropagation(definition);
+		return Flux.deferContextual((ctx) -> {
+			boolean createNewTransaction = handlePropagation(definition, ctx);
 
 			setOptionsFromDefinition(definition);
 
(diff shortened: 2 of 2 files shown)


spring-data-couchbase/src/main/java/org/springframework/data/couchbase/core/TransactionalSupport.java
```diff
@@ -50,6 +52,14 @@ public static Mono<Optional<CouchbaseResourceHolder>> checkForTransactionInThrea
 		});
 	}
 
+	public static Optional<CouchbaseResourceHolder> checkForTransactionInThreadLocalStorage(ContextView ctx) {
+		return Optional.ofNullable(ctx.hasKey(TransactionMarker.class) ? new CouchbaseResourceHolder(ctx.get(TransactionMarker.class).context()) : null);
```
programmatix, line 56: This isn't too great as a) it's not DRY or encapsulated - this functionality would really belong in TransactionMarkerOwner, and b) most importantly, it's removing a critical check.  We won't know if we're inside a blocking transaction here, as we don't have the ThreadLocal check.

spring-data-couchbase/src/main/java/org/springframework/data/couchbase/transaction/CouchbaseCallbackTransactionManager.java
```diff
@@ -187,8 +188,9 @@ public boolean isCompleted() {
 	}
 
 	// Propagation defines what happens when a @Transactional method is called from another @Transactional method.
-	private boolean handlePropagation(TransactionDefinition definition) {
-		boolean isExistingTransaction = TransactionalSupport.checkForTransactionInThreadLocalStorage().block().isPresent();
+	private boolean handlePropagation(TransactionDefinition definition, ContextView ctx) {
```
programmatix, line 191: What I meant on JCBC-1988 by "Refactor handlePropagation so we call TransactionalSupport.checkForTransactionInThreadLocalStorage() outside it and pass in the result."
is something like:

```
executeReactive():
		return TransactionalSupport.checkForTransactionInThreadLocalStorage().flatMap(isInTransaction -> {
		        boolean isInExistingTransaction = isInTransaction.isPresent();
			boolean createNewTransaction = handlePropagation(definition, isInExistingTransaction);
 
execute():
   boolean isInExistingTransaction = TransactionalSupport.checkForTransactionInThreadLocalStorage().block().isPresent();
   boolean createNewTransaction = handlePropagation(definition, isInExistingTransaction);
```
This preserves the full safety of the existing checkForTransactionInThreadLocalStorage() check (checking whether we are in either a reactive and blocking transaction), and just seems cleaner and simpler than the solution here.

spring-data-couchbase/src/main/java/org/springframework/data/couchbase/core/TransactionalSupport.java
mikereiche: > We won't know if we're inside a blocking transaction here, as we don't have the ThreadLocal check.
Both checks are there, however if the ctx is not due to the transaction, and there is indeed a blocking (ThreadLocal) transaction, we will miss it.  Hence the request for a TransactionMarkerOwner api that does require a ctx and return a Mono to get the ThreadLocal core.
```
	boolean isExistingTransaction = ctx != null ? TransactionalSupport.checkForTransactionInThreadLocalStorage(ctx).isPresent() :
				TransactionalSupport.checkForTransactionInThreadLocalStorage().block().isPresent();
```


Right.  So (a) requires the modification to TransactionMarkerOwner, and (b) needs what I've described in the other comment.

spring-data-couchbase/src/main/java/org/springframework/data/couchbase/transaction/CouchbaseCallbackTransactionManager.java
mikereiche: In the execute() path,  block() cannot be called as this may be in a NonBlocking thread.
TransactionMarkerOwner needs an api to get the Core out of ThreadLocal without calling block().

programmatix: The execute() path is inherently completely blocking (it is the blocking API) and hence can't be called from a NonBlocking thread.  If they're trying to do a blocking transaction inside a reactive operator, then that is an application bug.

programmatix: Did we get to the bottom of how they ended up on the execute() path from a reactive @Transactional?  I read through 1527 and I took from it that our initial guess was that they didn't do @EnableTransactionManagement, but in fact they had.

mikereiche: The springframework TransactionInterceptor was being used instead of CouchbaseTranasctionInterceptor because the customer was configuring using yaml for configuration and not a class that extends that AbstractCouchbaseConfiguration like we had been doing.  AbstractCouchbaseConfiguration has substitution of CouchbaseTransactionInterceptor.
I think you have a check for that in the code, but it hit the error on block() before the check.

PR #1546, opened 2022-08-26
Field Level Encryption Support.

### LINKED ISSUE #763: Field Level Encryption [DATACOUCH-455]
**[Domenico Ursoleo](https://jira.spring.io/secure/ViewProfile.jspa?name=udomenico)** opened **[DATACOUCH-455](https://jira.spring.io/browse/DATACOUCH-455?redirect=false)** and commented

Couchbase has a Field Level Encryption feature that encrypt field of a Document simply putting the `@EncryptField` annotation on a field of the document model and addind a CryptoManager to the environment,

With this configuration:

```java
@Configuration
@EnableCouchbaseRepositories(basePackages = { "com.example.demo.db" })
public class MyCouchbaseConfig extends AbstractCouchbaseConfiguration {
 
@Override 
protected CouchbaseEnvironment getEnvironment() { 
try {
 JceksKeyStoreProvider kp = new JceksKeyStoreProvider("secret");
 kp.publicKeyName("mypublickey");
 kp.storeKey("mypublickey",
           "!mysecretkey#9^5usdk39d&dlf)03sL".getBytes(Charset.forName("UTF-8")));
 kp.signingKeyName("HMACsecret");
 kp.storeKey("HMACsecret", "myauthpassword".getBytes(Charset.forName("UTF-8")));
 AES256CryptoProvider aes256CryptoProvider = new AES256CryptoProvider(kp);
 CryptoManager cryptoManager = new CryptoManager();
 cryptoManager.registerProvider("MyAESProvider", aes256CryptoProvider);
 return 
    DefaultCouchbaseEnvironment.builder().cryptoManager(cryptoManager).build();
 } catch (Exception ex) { return null; } }
}
```

 Given the model

 

```java
@Document
public class User {
@Id
@GeneratedValue(strategy = GenerationStrategy.UNIQUE)
private String id;

@Field("user")
private String userName;

@EncryptedField(provider = "MyAESProvider") 
@Field("pwd") 
private String password;

// Setters & getters
}
```

Persisting it via a simple CrudRepository should encrypt the password like calling the native couchbase java client:

```java
EntityDocument<User> document = EntityDocument.create(user);          bucket.repository().upsert(document);
 
```

 

 

---

**Affects:** 3.1.8 (Lovelace SR8)

1 votes, 2 watchers

mikereiche:
Note that an @Encrypted property object that contains an @Encrypted property is
not supported and will result in an exception.

Closes #763.

### DIFF
--- a/spring-data-couchbase/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/spring-data-couchbase/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -18,8 +18,9 @@ package org.springframework.data.couchbase.config;
 
 import static com.couchbase.client.java.ClusterOptions.clusterOptions;
 
-import java.util.Collections;
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.springframework.beans.factory.config.BeanDefinition;
@@ -27,6 +28,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Role;
+import org.springframework.core.convert.converter.GenericConverter;
 import org.springframework.core.type.filter.AnnotationTypeFilter;
 import org.springframework.data.convert.CustomConversions;
 import org.springframework.data.couchbase.CouchbaseClientFactory;
@@ -35,6 +37,7 @@ import org.springframework.data.couchbase.core.CouchbaseTemplate;
 import org.springframework.data.couchbase.core.ReactiveCouchbaseTemplate;
 import org.springframework.data.couchbase.core.convert.CouchbaseCustomConversions;
 import org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter;
+import org.springframework.data.couchbase.core.convert.OtherConverters;
 import org.springframework.data.couchbase.core.convert.translation.JacksonTranslationService;
 import org.springframework.data.couchbase.core.convert.translation.TranslationService;
 import org.springframework.data.couchbase.core.mapping.CouchbaseMappingContext;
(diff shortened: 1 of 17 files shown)


spring-data-couchbase/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -379,11 +400,28 @@ protected boolean autoIndexCreation() {
...
 	 * default.
 	 *
+	 * @param cryptoManagerOptional optional cryptoManager. Make varargs for backwards compatibility.
 	 * @return must not be {@literal null}.
 	 */
 	@Bean(name = BeanNames.COUCHBASE_CUSTOM_CONVERSIONS)
-	public CustomConversions customConversions() {
-		return new CouchbaseCustomConversions(Collections.emptyList());
+	public CustomConversions customConversions(CryptoManager... cryptoManagerOptional) {
+		assert (cryptoManagerOptional == null || cryptoManagerOptional.length <= 1);
```
dnault, line 408: What do you think of these alternatives stolen from StackOverflow?

```
public CustomConversions customConversions(
    Optional<CryptoManager> cryptoManagerOptional
)
```

Or this:

```
public CustomConversions customConversions(
    @Autowired(required = false) CryptoManager cryptoManager
)
```

Reference: https://stackoverflow.com/questions/46704950/how-to-define-optional-parameters-dependencies-in-the-bean-method-in-spring

mikereiche: The first one changes a public API so is not possible.
The second one - cryptoManager is not a bean and therefore cannot be instantiated with @Autowired. I also have doubts about using @Autowired in the configuration class.  And also bean-ifying things that don't absolutely need to be beans.

dnault: Changing `foo()` to `foo(T...)` is a source-compatible change, but I suspect it's not binary-compatible. (Maybe binary compatibility is not a concern?)

dnault (commented):
Looks like I missed the deadline -- sorry!

PR #1553, opened 2022-09-10
Refactor - Reorder modifier as per java spec

prasanth08:


mikereiche:
No specific ordering of class modifiers is required.

https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.1.1
"If two or more (distinct) class modifiers appear in a class declaration, then it is customary, though not required, that they appear in the order consistent with that shown above in the production for ClassModifier."

prasanth08:
I agree it's not required, There is a new Java Style Guidelines initiative in place for projects in the OpenJDK community. http://cr.openjdk.java.net/~alundblad/styleguide/index-v6.html#toc-modifiers 
Again it's not required but customary.

PR #1618, opened 2022-12-01
Add suport for JsonValue annotation on Enums.

### LINKED ISSUE #1617: @JsonValue ignored for enums
According to the docs of Couchbase's `@JsonValue`: *NOTE: when use for Java enums, one additional feature is that value returned by annotated method is also considered to be the value to deserialize from, not just JSON String to serialize as. This is possible since set of Enum values is constant and it is possible to define mapping, but can not be done in general for POJO types; as such, this is not used for POJO deserialization.*

The following enum should be deserialised / serialised using its `code` property. However, `@JsonValue` has no effect and Springdata Couch tries to **de**serialise by enum name. Note however that using the original Jackson annotation `@JsonValue`, the **de**serialisation does work !
In both cases, serialisation does not work, 

```
public enum ETurbulenceCategory {
    T10("10%"),
    T20("20%"),
    T30("30%");

    private final String code;

    ETurbulenceCategory(String code) {
        this.code = code;
    }
//  DOES WORK FOR DE-SERIALISATION, BUT NOT FOR SERIALSATION
    @com.fasterxml.jackson.annotation.JsonValue
//  DOES NOT WORK AT ALL
//  @com.couchbase.client.core.deps.com.fasterxml.jackson.annotation.JsonValue
    String getCode() {
        return code;
    }
}
```

mikereiche:
Implement the @ JsonValue annotation with a property-value converter.

Closes #1617.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
+++ b/src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
@@ -18,11 +18,16 @@ package org.springframework.data.couchbase.config;
 
 import static com.couchbase.client.java.ClusterOptions.clusterOptions;
 
+import java.lang.annotation.Annotation;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
+import com.couchbase.client.java.encryption.annotation.Encrypted;
+import com.fasterxml.jackson.annotation.JsonValue;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
@@ -37,9 +42,15 @@ import org.springframework.data.couchbase.CouchbaseClientFactory;
 import org.springframework.data.couchbase.SimpleCouchbaseClientFactory;
 import org.springframework.data.couchbase.core.CouchbaseTemplate;
 import org.springframework.data.couchbase.core.ReactiveCouchbaseTemplate;
+import org.springframework.data.couchbase.core.convert.BooleanToEnumConverterFactory;
 import org.springframework.data.couchbase.core.convert.CouchbaseCustomConversions;
 import org.springframework.data.couchbase.core.convert.CouchbasePropertyValueConverterFactory;
+import org.springframework.data.couchbase.core.convert.CryptoConverter;
+import org.springframework.data.couchbase.core.convert.IntegerToEnumConverterFactory;
+import org.springframework.data.couchbase.core.convert.JsonValueConverter;
 import org.springframework.data.couchbase.core.convert.MappingCouchbaseConverter;
+import org.springframework.data.couchbase.core.convert.OtherConverters;
+import org.springframework.data.couchbase.core.convert.StringToEnumConverterFactory;
 import org.springframework.data.couchbase.core.convert.translation.JacksonTranslationService;
 import org.springframework.data.couchbase.core.convert.translation.TranslationService;
 import org.springframework.data.couchbase.core.mapping.CouchbaseMappingContext;
@@ -60,7 +71,6 @@ import org.springframework.transaction.support.TransactionTemplate;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
-import com.couchbase.client.core.deps.com.fasterxml.jackson.databind.DeserializationFeature;
 import com.couchbase.client.core.encryption.CryptoManager;
 import com.couchbase.client.core.env.Authenticator;
 import com.couchbase.client.core.env.PasswordAuthenticator;
@@ -72,6 +82,7 @@ import com.couchbase.client.java.env.ClusterEnvironment;
 import com.couchbase.client.java.json.JacksonTransformers;
 import com.couchbase.client.java.json.JsonValueModule;
 import com.couchbase.client.java.query.QueryScanConsistency;
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 /**
@@ -87,8 +98,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 @Configuration
 public abstract class AbstractCouchbaseConfiguration {
 
-	ObjectMapper mapper;
-	CryptoManager cryptoManager = null;
+	volatile ObjectMapper objectMapper;
+	volatile CryptoManager cryptoManager = null;
 
 	/**
 	 * The connection string which allows the SDK to connect to the cluster.
@@ -157,7 +168,7 @@ public abstract class AbstractCouchbaseConfiguration {
 		if (!nonShadowedJacksonPresent()) {
 			throw new CouchbaseException("non-shadowed Jackson not present");
 		}
-		builder.jsonSerializer(JacksonJsonSerializer.create(getCouchbaseObjectMapper()));
+		builder.jsonSerializer(JacksonJsonSerializer.create(getObjectMapper()));
 		builder.cryptoManager(getCryptoManager());
 		configureEnvironment(builder);
 		return builder.build();
@@ -277,10 +288,12 @@ public abstract class AbstractCouchbaseConfiguration {
 	@Bean
 	public TranslationService couchbaseTranslationService() {
 		final JacksonTranslationService jacksonTranslationService = new JacksonTranslationService();
-		jacksonTranslationService.setObjectMapper(getCouchbaseObjectMapper());
+		jacksonTranslationService.setObjectMapper(getObjectMapper());
 		jacksonTranslationService.afterPropertiesSet();
 		// for sdk3, we need to ask the mapper _it_ uses to ignore extra fields...
-		JacksonTransformers.MAPPER.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+		JacksonTransformers.MAPPER.configure(
+				com.couchbase.client.core.deps.com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES,
+				false);
 		return jacksonTranslationService;
 	}
 
@@ -298,21 +311,26 @@ public abstract class AbstractCouchbaseConfiguration {
 		return mappingContext;
 	}
 
-	private ObjectMapper getCouchbaseObjectMapper() {
-		if (mapper != null) {
-			return mapper;
+	final public ObjectMapper getObjectMapper() {
+		if(objectMapper == null) {
+			synchronized (this) {
+				if (objectMapper == null) {
+					objectMapper = couchbaseObjectMapper();
+				}
+			}
 		}
-		return mapper = couchbaseObjectMapper();
+		return objectMapper;
 	}
 
 	/**
-	 * Creates a {@link ObjectMapper} for the jsonSerializer of the ClusterEnvironment
+	 * Creates a {@link ObjectMapper} for the jsonSerializer of the ClusterEnvironment and spring-data-couchbase
+	 * jacksonTranslationService and also some converters (EnumToObject, StringToEnum, IntegerToEnum)
 	 *
 	 * @return ObjectMapper
 	 */
-	public ObjectMapper couchbaseObjectMapper() {
-		ObjectMapper om = new ObjectMapper(); // or use the one from the Java SDK (?) JacksonTransformers.MAPPER
-		om.configure(com.fasterxml.jackson.databind.DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+	protected ObjectMapper couchbaseObjectMapper() {
+		ObjectMapper om = new ObjectMapper();
+		om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 		om.registerModule(new JsonValueModule());
 		if (getCryptoManager() != null) {
 			om.registerModule(new EncryptionModule(getCryptoManager()));
@@ -400,20 +418,35 @@ public abstract class AbstractCouchbaseConfiguration {
 		List<GenericConverter> newConverters = new ArrayList();
 		CustomConversions customConversions = CouchbaseCustomConversions.create(configurationAdapter -> {
 			SimplePropertyValueConversions valueConversions = new SimplePropertyValueConversions();
-			valueConversions.setConverterFactory(new CouchbasePropertyValueConverterFactory(cryptoManager));
+			valueConversions.setConverterFactory(new CouchbasePropertyValueConverterFactory(cryptoManager, annotationToConverterMap()));
 			valueConversions.setValueConverterRegistry(new PropertyValueConverterRegistrar().buildRegistry());
+			valueConversions.afterPropertiesSet(); // wraps the CouchbasePropertyValueConverterFactory with CachingPVCFactory
 			configurationAdapter.setPropertyValueConversions(valueConversions);
 			configurationAdapter.registerConverters(newConverters);
+			configurationAdapter.registerConverter(new OtherConverters.EnumToObject(getObjectMapper()));
+			configurationAdapter.registerConverterFactory(new IntegerToEnumConverterFactory(getObjectMapper()));
+			configurationAdapter.registerConverterFactory(new StringToEnumConverterFactory(getObjectMapper()));
+			configurationAdapter.registerConverterFactory(new BooleanToEnumConverterFactory(getObjectMapper()));
 		});
 		return customConversions;
 	}
 
+	Map<Class<? extends Annotation>,Class<?>> annotationToConverterMap(){
+		Map<Class<? extends Annotation>,Class<?>> map= new HashMap();
+		map.put(Encrypted.class, CryptoConverter.class);
+		map.put(JsonValue.class, JsonValueConverter.class);
+		return map;
+	}
 	/**
 	 * cryptoManager can be null, so it cannot be a bean and then used as an arg for bean methods
 	 */
 	private CryptoManager getCryptoManager() {
-		if (cryptoManager == null) {
-			cryptoManager = cryptoManager();
+		if(cryptoManager == null) {
+			synchronized (this) {
+				if (cryptoManager == null) {
+					cryptoManager = cryptoManager();
+				}
+			}
 		}
 		return cryptoManager;
 	}
--- a/src/main/java/org/springframework/data/couchbase/core/convert/AbstractCouchbaseConverter.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/AbstractCouchbaseConverter.java
@@ -117,10 +117,8 @@ public abstract class AbstractCouchbaseConverter implements CouchbaseConverter, 
 			return null;
 		}
 		if (processValueConverter && conversions.hasValueConverter(prop)) {
-			CouchbaseDocument encrypted = (CouchbaseDocument) conversions.getPropertyValueConversions()
-					.getValueConverter(prop)
-					.write(value, new CouchbaseConversionContext(prop, (MappingCouchbaseConverter) this, accessor));
-			return encrypted;
+			return conversions.getPropertyValueConversions().getValueConverter(prop).write(value,
+					new CouchbaseConversionContext(prop, (MappingCouchbaseConverter) this, accessor));
 		}
 		Class<?> targetClass = this.conversions.getCustomWriteTarget(value.getClass()).orElse(null);
 
@@ -134,7 +132,9 @@ public abstract class AbstractCouchbaseConverter implements CouchbaseConverter, 
 		Object result = this.conversions.getCustomWriteTarget(prop.getType()) //
 				.map(it -> this.conversionService.convert(value, new TypeDescriptor(prop.getField()),
 						TypeDescriptor.valueOf(it))) //
-				.orElseGet(() -> Enum.class.isAssignableFrom(value.getClass()) ? ((Enum<?>) value).name() : value);
+				.orElse(value);
+		// superseded by Enum converters
+		// .orElseGet(() -> Enum.class.isAssignableFrom(value.getClass()) ? ((Enum<?>) value).name() : value);
 
 		return result;
 
@@ -160,7 +160,7 @@ public abstract class AbstractCouchbaseConverter implements CouchbaseConverter, 
 		Class<?> elementType = value.getClass();
 
 		if (elementType == null || conversions.isSimpleType(elementType)) {
-			value = Enum.class.isAssignableFrom(value.getClass()) ? ((Enum<?>) value).name() : value;
+			// superseded by EnumCvtrs value = Enum.class.isAssignableFrom(value.getClass()) ? ((Enum<?>) value).name() : value;
 		} else if (value instanceof Collection || elementType.isArray()) {
 			TypeInformation<?> type = ClassTypeInformation.from(value.getClass());
 			value = ((MappingCouchbaseConverter) this).writeCollectionInternal(MappingCouchbaseConverter.asCollection(value),
--- a/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseCustomConversions.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseCustomConversions.java
@@ -44,8 +44,6 @@ import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.model.SimpleTypeHolder;
 import org.springframework.util.Assert;
 
-import com.couchbase.client.java.encryption.annotation.Encrypted;
-
 /**
  * Value object to capture custom conversion.
  * <p>
@@ -112,9 +110,6 @@ public class CouchbaseCustomConversions extends org.springframework.data.convert
 
 	@Override
 	public boolean hasValueConverter(PersistentProperty<?> property) {
-		if (property.findAnnotation(Encrypted.class) != null) {
-			return true;
-		}
 		return super.hasValueConverter(property);
 	}
 
--- a/src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors
+ * Copyright 2022 the original author or authors
  *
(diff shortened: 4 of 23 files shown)


src/main/java/org/springframework/data/couchbase/core/convert/AbstractCouchbaseConverter.java
```diff
@@ -117,10 +117,8 @@ public Object convertForWriteIfNeeded(CouchbasePersistentProperty prop, Converti
 			return null;
 		}
 		if (processValueConverter && conversions.hasValueConverter(prop)) {
-			CouchbaseDocument encrypted = (CouchbaseDocument) conversions.getPropertyValueConversions()
-					.getValueConverter(prop)
-					.write(value, new CouchbaseConversionContext(prop, (MappingCouchbaseConverter) this, accessor));
-			return encrypted;
+			return conversions.getPropertyValueConversions().getValueConverter(prop).write(value,
+					new CouchbaseConversionContext(prop, (MappingCouchbaseConverter) this, accessor));
 		}
```
mikereiche, line 122: The converted element can be anything.

src/main/java/org/springframework/data/couchbase/core/convert/CouchbaseCustomConversions.java
```diff
@@ -112,9 +110,6 @@ public static CouchbaseCustomConversions create(Consumer<CouchbaseConverterConfi
 
 	@Override
 	public boolean hasValueConverter(PersistentProperty<?> property) {
-		if (property.findAnnotation(Encrypted.class) != null) {
-			return true;
-		}
 		return super.hasValueConverter(property);
```
mikereiche, line 113: super.hasValueConverter() eventually calls CouchbasePropertyValueConverterFactory.getConverter() which does this check.

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -18,26 +18,28 @@
...
+ * Accept the Couchbase @Encrypted annotation in addition to @ValueConverter.<br>
+ * Note that valueConversions.afterPropertiesSet() encapsulates this in a CachingPropertyValueConverterFactory making
+ * caching in this class unnecessary.
  *
  * @author Michael Reiche
  */
 public class CouchbasePropertyValueConverterFactory implements PropertyValueConverterFactory {
 
 	CryptoManager cryptoManager;
-	Map<Class<? extends PropertyValueConverter<?, ?, ?>>, PropertyValueConverter<?, ?, ?>> converterCache = new HashMap<>();
```
mikereiche, line 43: don't need caching in our converter factory as it is encapsulated in a CachingPropertyValueConverterFactory

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -50,35 +52,48 @@ public <DV, SV, P extends ValueConversionContext<?>> PropertyValueConverter<DV,
...
 		}
+
+		// for Enums only
+
+		for (Method m : type.getDeclaredMethods()) {
+			if (m.getAnnotation(JsonValue.class) != null) {
+				Class jsonValueConverterClass = JsonValueConverter.class;
+				return getConverter((Class<PropertyValueConverter<DV, SV, P>>) jsonValueConverterClass);
+			}
+		}
```
mikereiche, line 74: look for a JsonValue annotation. 
@dnault -  Should this also look for a shaded JsonValue annotation?

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -50,35 +52,48 @@ public <DV, SV, P extends ValueConversionContext<?>> PropertyValueConverter<DV,
...
 
 	@Override
 	public <DV, SV, P extends ValueConversionContext<?>> PropertyValueConverter<DV, SV, P> getConverter(
 			Class<? extends PropertyValueConverter<DV, SV, P>> converterType) {
 
-		PropertyValueConverter<?, ?, ?> converter = converterCache.get(converterType);
-		if (converter != null) {
-			return (PropertyValueConverter<DV, SV, P>) converter;
-		}
+		PropertyValueConverter<?, ?, ?> converter = null;
```
mikereiche, line 83: cache here would be redundant

src/main/java/org/springframework/data/couchbase/core/convert/translation/JsonValueConverter.java
```diff
@@ -0,0 +1,112 @@
...
+		implements PropertyValueConverter<Enum<?>, Object, ValueConversionContext<? extends PersistentProperty<?>>> {
+
+	Map<Class<? extends Enum>, Method> methodCache = new HashMap();
+	Map<Class<? extends Enum>, Map<Object, Enum<?>>> enumConstantsCache = new HashMap();
+
+	@Override
+	public Enum<?> read(Object value, ValueConversionContext<? extends PersistentProperty<?>> context) {
+		Class<Enum> type = (Class<Enum>) context.getProperty().getActualType();
+		return getEnumConstant(type, value);
+	}
```
mikereiche, line 44: From the number or string, return the enum constant from whence it came.

src/main/java/org/springframework/data/couchbase/core/convert/translation/JsonValueConverter.java
```diff
@@ -0,0 +1,112 @@
...
+
+	@Override
+	public Object write(Enum<?> value, ValueConversionContext<? extends PersistentProperty<?>> context) {
+		try {
+			Class<? extends Enum> type = value.getClass();
+			return getJsonValueMethod(type).invoke(value);
+		} catch (IllegalAccessException | InvocationTargetException e) {
+			throw new RuntimeException(e);
+		}
+	}
```
mikereiche, line 54: write the result of the annotated method.

src/main/java/org/springframework/data/couchbase/core/convert/translation/JsonValueConverter.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private Enum<?> getEnumConstant(Class<Enum> type, Object value) {
+		Map<Object, Enum<?>> enumConstants = enumConstantsCache.get(type);
```
mikereiche, line 57: check for cached mapping of enumConstants for this Enum.

src/main/java/org/springframework/data/couchbase/core/convert/translation/JsonValueConverter.java
```diff
@@ -0,0 +1,112 @@
...
+			enumConstantsCache.put(type, enumConstants);
+			Method jsonValueMethod = getJsonValueMethod(type);
+			for (Enum<?> e : type.getEnumConstants()) { // cache all the constants while we're here
+				try {
+					enumConstants.put(jsonValueMethod.invoke(e), e);
+				} catch (IllegalAccessException | InvocationTargetException ex) {
+					throw new RuntimeException(ex);
+				}
+			}
+		}
```
mikereiche, line 69: build cache for all the @JsonValues -> Enum constants for this enum.

src/main/java/org/springframework/data/couchbase/core/convert/translation/JsonValueConverter.java
```diff
@@ -0,0 +1,112 @@
...
+		if (jsonValueMethod != null) {
+			return jsonValueMethod;
+		}
+		for (Method m : type.getDeclaredMethods()) {
+			JsonValue jsonValueAnn = m.getAnnotation(JsonValue.class);
+			if (jsonValueAnn != null && jsonValueAnn.value()) {
+				jsonValueMethod = m;
+				break;
+			}
+		}
```
mikereiche, line 89: find the annotated method that is enabled (value() == true).

src/main/java/org/springframework/data/couchbase/core/convert/translation/JsonValueConverter.java
```diff
@@ -0,0 +1,112 @@
...
+				}
+			}
+		}
+		// if nothing was found, use toString()
+		if (jsonValueMethod == null) {
+			try {
+				jsonValueMethod = type.getMethod("toString");
+			} catch (NoSuchMethodException e) {
+				throw new RuntimeException(e);
+			}
```
mikereiche, line 105: If there was no enabled @ JsonValue method, instead use toString().  This will be ok for serialization, but deserialization will fail.  Maybe we should just throw an exception, as it is impossible to get here without the converter factory finding an @JsonValue method that is enabled.

src/main/java/org/springframework/data/couchbase/core/convert/MappingCouchbaseConverter.java
```diff
@@ -954,18 +955,25 @@ private <R> R readValue(Object value, TypeInformation type, Object parent) {
...
 			return (R) read(prop.getTypeInformation(), (CouchbaseDocument) value, parent);
-		} else if (value instanceof CouchbaseList) {
+		}
+		if (value instanceof CouchbaseList) {
 			return (R) readCollection(prop.getTypeInformation(), (CouchbaseList) value, parent);
-		} else {
-			return (R) getPotentiallyConvertedSimpleRead(value, prop);// passes PersistentProperty with annotations
 		}
+		return (R) getPotentiallyConvertedSimpleRead(value, prop);// passes PersistentProperty with annotations
+
```
mikereiche, line 976: This changes allows the read-attempts to fall-through.

src/test/java/org/springframework/data/couchbase/domain/AirportJsonValuedObject.java
```diff
@@ -0,0 +1,39 @@
...
+
+	@PersistenceCreator
+	@JsonCreator
+	public AirportJsonValuedObject(String theValue) {
+		this.theValue = reverseMap.get(theValue);
+		if (this.theValue == null) {
+			throw new RuntimeException(
+					"invalid value for " + this.getClass().getName() + " : " + theValue + ". Must be one of " + reverseMap);
+		}
+	}
```
mikereiche, line 24: This is the equivalent of creating an Enum from the output of its @JsonValue method.

src/main/java/org/springframework/data/couchbase/core/convert/JsonValueConverter.java
```diff
@@ -0,0 +1,85 @@
...
+ * Enum properties annotated with JsonValue
+ * 
+ * @author Michael Reiche
+ */
+public class JsonValueConverter
+		implements PropertyValueConverter<Object, Object, ValueConversionContext<? extends PersistentProperty<?>>> {
+
+	static protected Map<Class<?>, Method> valueMethodCache = new ConcurrentHashMap<>();
+	static protected Map<Class<?>, Constructor<?>> creatorMethodCache = new ConcurrentHashMap<>();
+	static protected Map<Class<?>, Map<Object, Enum<?>>> enumConstantsCache = new ConcurrentHashMap<>();
```
dnault, line 38: nit: should these maps all be final?

src/main/java/org/springframework/data/couchbase/core/convert/JsonValueConverter.java
```diff
@@ -0,0 +1,85 @@
...
+			}
+		}
+
+		// for Enums, we can use the inverse mapping we made from the @JsonValue method
+		if (Enum.class.isAssignableFrom(type)) {
+			return getEnumConstant((Class<Enum<?>>) type, value);
+		}
+
+		// Otherwise fall-through in MappingCouchbaseConverter.readValue()
+		throw new ConverterHasNoConversion();
```
dnault, line 59: Filling in an exception's stack trace is relatively expensive. Could this potentially happen every time SDC deserializes an object?  Possible mitigation is to create a single instance of the exception, store it in a static final, and throw that instance over and over again.

src/test/java/org/springframework/data/couchbase/domain/Airport.java
```diff
@@ -48,8 +52,16 @@ public class Airport extends ComparableEntity {
...
+	public AirportJsonValuedObject jvObject;
+
+	public Airport() {
+		this.jvObject = new AirportJsonValuedObject();
+		this.jvObject.theValue = "first";
+	}
+
 	@PersistenceConstructor
 	public Airport(String key, String iata, String icao) {
+		this();
```
dnault, line 64: NIT: A more reliable way to share init code between constructors is to put the code in a non-static init block.

```
{
  this.jvObject = new AirportJsonValuedObject();
  this.jvObject.theValue = "first";
}
```

https://stackoverflow.com/questions/3987428/what-is-an-initialization-block

src/test/java/org/springframework/data/couchbase/domain/AirportJsonValuedObject.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.data.annotation.PersistenceCreator;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public class AirportJsonValuedObject {
+
+	public AirportJsonValuedObject() {}
+
+	@PersistenceCreator
+	@JsonCreator
```
dnault, line 17: How is this new feature related to `@JsonCreater`? Shouldn't Spring Data code use `@PersistenceCreator` instead?

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -16,70 +16,171 @@
...
+		for (Annotation ann : property.getField().getAnnotations()) {
+			Class<?> converterClass = converterFromFieldAnnotation(ann);
+			if (converterClass != null) {
+				return getConverter((Class<PropertyValueConverter<DV, SV, P>>) converterClass, property);
+			}
+		}
+		Optional<PropertyValueConverter<?, ?, ?>> cachedConverterForType = converterCacheForType
+				.get(property.getActualType());
+		if (cachedConverterForType != null) {
+			return (PropertyValueConverter<DV, SV, P>) cachedConverterForType.get();
```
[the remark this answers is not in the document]
mikereiche: yes. but it was scary.  Also made me see my cache was broken. 

return (PropertyValueConverter<DV, SV, P> )converterCacheForType
				.computeIfAbsent(property.getType(), p -> 
                                Optional.ofNullable((maybeTypePropertyConverter(property)))).orElse(null);

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -16,70 +16,171 @@
...
+				// for Enums, we can compute the conversion on read()
+				// otherwise, we rely on a constructor with an argument that is the output of the JsonValue method (inverse)
+				if (Enum.class.isAssignableFrom(property.getActualType())) {
+					Map<Object, Enum<?>> enumConstants = new HashMap<>();
+					JsonValueConverter.enumConstantsCache.put((Class<Enum<?>>) type, enumConstants);
+					Class<Enum<?>> enumType = (Class<Enum<?>>) type;
+					for (Enum<?> e : enumType.getEnumConstants()) { // create a value -> enum-object map for Enums
+						try {
+							enumConstants.put(jsonValueMethod.invoke(e), e);
+						} catch (IllegalAccessException | InvocationTargetException ex) {
```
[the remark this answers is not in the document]
mikereiche: The IllegalAccessException should never occur.  The InvocationTargetException can.

src/main/java/org/springframework/data/couchbase/core/convert/JsonValueConverter.java
mikereiche: yes.

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -16,70 +16,171 @@
...
+			JsonValue jsonValueAnn = m.getAnnotation(JsonValue.class);
+			if (jsonValueAnn != null && jsonValueAnn.value()) {
+				jsonValueMethod = m;
+				jsonValueMethod.setAccessible(true);
+				JsonValueConverter.valueMethodCache.put(type, jsonValueMethod);
+				// for Enums, we can compute the conversion on read()
+				// otherwise, we rely on a constructor with an argument that is the output of the JsonValue method (inverse)
+				if (Enum.class.isAssignableFrom(property.getActualType())) {
+					Map<Object, Enum<?>> enumConstants = new HashMap<>();
+					JsonValueConverter.enumConstantsCache.put((Class<Enum<?>>) type, enumConstants);
```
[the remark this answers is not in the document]
mikereiche: yes

src/test/java/org/springframework/data/couchbase/domain/Airport.java
mikereiche: yes

src/test/java/org/springframework/data/couchbase/domain/AirportJsonValuedObject.java
mikereiche: Aside from my hack with storing the enum constants in a map, there should aways be an @JsonCreator method if one wants to deserialize an object from what the output serialized by the @JsonValue.  What I have in the JsonValueConverter.read(), if the method can't figure out how to deserialize with either an @JsonCreator method or the enum constants hack, it throws an exception which is caught by the calling readValue() which goes on to deserialize by the regular spring-data process.  It first looks for an @PersistenceCreator constructor to use. (If it doesn't find one, it creates the object with a no-args constructor and then sets the args).  
I used this for testing both, first one, then the other manually.  When both are present, the JsonValueConverter gets the first try - with the @JsonCreator.  I should have made two objects, one with @PersistenceCreateor, one with @JsonCreator.

src/main/java/org/springframework/data/couchbase/core/convert/StringToEnumConverterFactory.java
```diff
@@ -0,0 +1,83 @@
...
+		}
+
+		@Override
+		@Nullable
+		public T convert(String source) {
+			if (source == null) {
+				return null;
+			}
+			try {
+				return objectMapper.readValue("\"" + source + "\"", enumType);
```
dnault, line 76: This approach has string escaping issues; just adding quotes is not sufficient for turning a string into JSON. More robust approach:

    String jsonString = objectMapper.writeValueAsString(source);
    return objectMapper.readValue(jsonString, enumType);

Better yet, skip the intermediate step and let Jackson convert from String directly to Enum:

    objectMapper.convertValue(source, enumType);

src/test/java/org/springframework/data/couchbase/domain/Airport.java
```diff
@@ -39,15 +43,30 @@ public class Airport extends ComparableEntity {
...
 	@CreatedBy private String createdBy;
 	@Expiration private long expiration;
 	@Max(2) long size;
 	private long someNumber;
 
+	public AirportJsonValuedObject jvObject;
+
+	{
+		this.jvObject = new AirportJsonValuedObject();
+		this.jvObject.theValue="first";
```
dnault, line 64: NIT: formatting?

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
```diff
@@ -306,20 +311,30 @@ private ObjectMapper getCouchbaseObjectMapper() {
...
 		om.registerModule(new JsonValueModule());
 		if (getCryptoManager() != null) {
 			om.registerModule(new EncryptionModule(getCryptoManager()));
 		}
 		return om;
 	}
 
+	ObjectMapper objectMapper;
+
+	ObjectMapper getObjectMapper() {
```
dnault, line 331: A few concerns:

* `couchbaseObjectMapper` and `getObjectMapper` do not appear to be thread-safe. Is that okay?

* Beacuse `getObjectMapper` caches the instance, `couchbaseObjectMapper` registers additional modules each time it is called.

* Again, because the instance is cached, the ObjectMapper returned by `getObjectMapper` might or might not have the configuration that was applied by couchbaseObjectMapper().

ObjectMapper is considered "thread-safe after configuration". It's not obvious from looking at the code that all calls to `couchbaseObjectMapper()` (which configures the mapper) happen before the mapper is used.


Suggestion:

* Remove the getObjectMapper() method, and call couchbaseObjectMapper() instead. Make couchbaseObjectMapper() responsible for caching the instance.

* Make the `objectMapper` field volatile.

* Inside `couchbaseObjectMapper()`, use double-checked locking to initialize the field if necessary. (This idiom is safe if the field being intialized is volatile.)
```
if (objectMapper == null) {
    synchronized {
        if (objectMapper == null) {
            objectMapper = new ObjectMapper();
            // configure it
        }
    }
}
```

src/test/java/org/springframework/data/couchbase/domain/AirportJsonValuedObject.java
dnault: Got it! Thanks for the explanation.

dnault (changes requested):

src/main/java/org/springframework/data/couchbase/core/convert/IntegerToEnumConverterFactory.java
```diff
@@ -0,0 +1,83 @@
...
+
+	@Override
+	public <T extends Enum> Converter<Integer, T> getConverter(Class<T> targetType) {
+		return new ObjectToEnum(getEnumType(targetType), objectMapper);
+	}
+
+	public static Class<?> getEnumType(Class<?> targetType) {
+		Class<?> enumType = targetType;
+		while (enumType != null && !enumType.isEnum()) {
+			enumType = enumType.getSuperclass();
```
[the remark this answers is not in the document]
mikereiche: The two classes are stolen from classes of the same name - https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/core/convert/support/StringToEnumConverterFactory.java

The getEnumType() method is from :

https://github.com/spring-projects/spring-framework/blob/6e08c56076280021542ee086a6ddd971a256c5ca/spring-core/src/main/java/org/springframework/core/convert/support/ConversionUtils.java#L74

src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
```diff
@@ -187,4 +197,36 @@ public Class<?> convert(String source) {
...
+	/**
+	 * Writing converter for Enums. This is registered in
+	 * {@link org.springframework.data.couchbase.config.AbstractCouchbaseConfiguration#customConversions( CryptoManager)}.
+	 * The corresponding reading converters are in {@link IntegerToEnumConverterFactory} and
+	 * {@link StringToEnumConverterFactory}
+	 */
+
+	@WritingConverter
+	public static class EnumToObject implements Converter<Enum<?>, Object> {
+		ObjectMapper objectMapper;
```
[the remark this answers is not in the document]
mikereiche: no.  AbstractCouchbaseConfiguration.java:431
			configurationAdapter.registerConverter(new OtherConverters.EnumToObject(getObjectMapper()));
			
I did not figure out how to pass in this objectMapper into this block in CouchbaseCustomConversions where the static OtherConverters.getConvertersToRegister()) is called, otherwise it could be private.  (Also, all the other converters in OtherConverters are 'public' (although they do not need to be).)

	static {

		List<Object> converters = new ArrayList<>();

		converters.addAll(DateConverters.getConvertersToRegister());
		converters.addAll(CouchbaseJsr310Converters.getConvertersToRegister());
		converters.addAll(OtherConverters.getConvertersToRegister());

		STORE_CONVERTERS = Collections.unmodifiableList(converters);
		STORE_CONVERSIONS = StoreConversions.of(SimpleTypeHolder.DEFAULT, STORE_CONVERTERS);
	}

src/main/java/org/springframework/data/couchbase/core/convert/StringToEnumConverterFactory.java
mikereiche: Thanks. Much better.

src/main/java/org/springframework/data/couchbase/config/AbstractCouchbaseConfiguration.java
mikereiche: Something went off the rails here.   There was already a mapper object for caching.  Fix is coming.

The intention is to separate the caching from the initialization - as the user can override the initialization, and I don't want to burden them (or trust them) with the caching.  This would normally be an @Bean in the configuration, but it cannot be (there's a previous issue for that).  It is similar to cryptoManager which also cannot be an @Bean (because it can be null). 

 It should follow the same pattern as cryptoManager : getCryptoManager()  provides the cached cryptoManager, if necessary calling cryptoManager() which the user can override to create the cryptoManager.

So getObjectManager() provides the cached objectMapper, if necessary calling couchbaseObjectMapper() which the user may override for customization.

couchbaseObjectMapper() should be called only once.  I would make it private, but a user's override may want to call super.couchbaseObjectMapper(). (this is actually in one of the test cases).

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
```diff
@@ -16,70 +16,171 @@
...
+					Class<Enum<?>> enumType = (Class<Enum<?>>) type;
+					for (Enum<?> e : enumType.getEnumConstants()) { // create a value -> enum-object map for Enums
+						try {
+							enumConstants.put(jsonValueMethod.invoke(e), e);
```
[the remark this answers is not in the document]
mikereiche: (1) I eliminated this by just deferring to ObjectMapper.
(2)But ObjectMapper does the same thing as I was doing.

src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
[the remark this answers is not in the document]
mikereiche: I misunderstood.  I thought the comment was for the method on the line above instead of the object indicated. Fixed.

src/test/java/org/springframework/data/couchbase/domain/Airport.java
mikereiche: fixed

src/main/java/org/springframework/data/couchbase/core/convert/CouchbasePropertyValueConverterFactory.java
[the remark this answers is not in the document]
mikereiche: removed enumConstantsCache

PR #1633, opened 2023-01-03
Use "= true" for TRUE() and "= false" for FALSE().

### LINKED ISSUE #1627: QueryCriteria TRUE() to N1QL is behaving differently in version 4.4.5
The SpEl query extraction/preparation is behaving wrong after upgrade from **spring-data-couchbase:3.2.6.RELESE** to spring-**data-couchbase:4.4.5**.  The keyword IsTrue/True is expected to be transformed as 'field = TRUE', but it does not. Instead it just adds the field to the query as shown below.

This causes queries perform badly since they don’t pick expected indexes plus possible invalid results.

**Before** : 
Optional<UserDocument> findByIdAndStatusAnd**ActiveDocumentTrue**(String id, String status);

query: SELECT `_class`, `id`, META(`bucket`).`id` AS __id,…. FROM `bucket` WHERE `_class` = "com.blah.UserDocument" AND `id` = $1 and `status` = $2 and **_`activeDocument` = true_**

**After**:
Optional<UserDocument> findByIdAndStatusAnd**ActiveDocumentTrue**(String id, String status);

query: SELECT `_class`, `id`, META(`bucket`).`id` AS __id,…. FROM `bucket` WHERE `_class` = "com.blah.UserDocument" AND `id` = $1 and `status` = $2 and **_`activeDocument`_**

mikereiche:
Closes #1627.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/query/QueryCriteria.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/QueryCriteria.java
@@ -499,14 +499,14 @@ public class QueryCriteria implements QueryCriteriaDefinition {
 	public QueryCriteria TRUE() { // true/false are reserved, use TRUE/FALSE
 		value = null;
 		operator = null;
-		format = "%1$s"; // field = 1$, operator = 2$, value=$3, $4, ...
+		format = "%1$s = true"; // field = 1$, operator = 2$, value=$3, $4, ...
 		return this;
 	}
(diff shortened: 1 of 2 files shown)


mikereiche (commented):
@dnault - field = true and field = false has a slightly different behavior than using just field and NOT(field)

dnault (approved):

PR #1640, opened 2023-01-10
Take expiry from entity class annotation for getAndTouch.

### LINKED ISSUE #1634: Never call getAndTouch in ReactiveFindByIdOperationSupport
Hello

I'm trying to do a **getAndTouch** in the
https://github.com/spring-projects/spring-data-couchbase/blob/d994eb7bb3cf15f2ede1acc84692316c9f732958/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByIdOperationSupport.java#L88-L90

method and I think there is a bug.
The options are never initialized so the get method is always executed.

Related issue: https://github.com/spring-projects/spring-data-couchbase/pull/1196

Thank you very much

mikereiche:
Closes #1634.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByIdOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveFindByIdOperationSupport.java
@@ -72,6 +72,8 @@ public class ReactiveFindByIdOperationSupport implements ReactiveFindByIdOperati
 		private final ReactiveTemplateSupport support;
 		private final Duration expiry;
 
+		private Duration expiryToUse;
+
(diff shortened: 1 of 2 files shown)


mikereiche (commented):
@jorgerod @dnault

src/main/java/org/springframework/data/couchbase/core/ReactiveFindByIdOperationSupport.java
```diff
@@ -72,6 +72,8 @@ public <T> ReactiveFindById<T> findById(Class<T> domainType) {
 		private final ReactiveTemplateSupport support;
 		private final Duration expiry;
 
+		private Duration expiryToUse;
```
dnault, line 75: Does this need to be volatile?

dnault (approved):

PR #1649, opened 2023-01-20
Durability via Annotations

### LINKED ISSUE #1486: Support PersistTo, ReplicateTo and DurabilityLevel attributes on Document annotation.
Hello

In some operationSupport, the values of PersistTo, ReplicateTo or DurabilityLevel are not configurable and the value NONE is always being used.

https://github.com/spring-projects/spring-data-couchbase/blob/34f96018350cb1d1d556d56981df738749bf59a7/src/main/java/org/springframework/data/couchbase/core/ReactiveReplaceByIdOperationSupport.java#L45-L50

https://github.com/spring-projects/spring-data-couchbase/blob/62feffa89d2264e1fdfdb9913703f4f106a7d60b/src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByIdOperationSupport.java#L49-L54

Is there any plan to support these values?

Thank you very much in advance

babltiga:
Added support of the PersistTo, ReplicateTo, and DurabilityLevel via @Document annotation.

Closes #1486

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
@@ -31,6 +31,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.data.couchbase.core.mapping.CouchbaseDocument;
+import org.springframework.data.couchbase.core.mapping.Document;
+import org.springframework.data.couchbase.core.mapping.Expiry;
 import org.springframework.data.couchbase.repository.Collection;
 import org.springframework.data.couchbase.repository.ScanConsistency;
 import org.springframework.data.couchbase.repository.Scope;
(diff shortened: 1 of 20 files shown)


mikereiche:
Thanks. I will take a look at it.

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -225,6 +228,30 @@ public static String getScopeFrom(Class<?> domainType) {
...
+		return document.persistTo();
+	}
+
+	public static ReplicateTo getReplicateTo(Class<?> domainType) {
+		if (domainType == null) {
+			return null;
+		}
+		Document document = AnnotatedElementUtils.findMergedAnnotation(domainType, Document.class);
+		return document.replicateTo();
+	}
```
mikereiche, line 254: Just two minor changes in each method :

1) The options cannot be null as they get set if they are not NONE (see buildRemoveOptions) and null is not allowed.

2) it's possible for document to be null, so also return NONE.

Also - to run all the integration tests against a couchbase server, replace src/test/resources/integration.properties with one like this:

```
# Options: containerized, mocked, unmanaged
cluster.type=unmanaged
# Entry point configuration if not managed
# value of hostname and ns_server port
cluster.unmanaged.seed=127.0.0.1:8091
cluster.adminUsername=Administrator
cluster.adminPassword=password
cluster.unmanaged.numReplicas=0
```

and then `mvn integration-test`

Issue #1662, opened 2023-02-11
 ReactiveCouchbaseTemplateKeyValueIntegrationTests.findByIdWithLock  needs longer duration


mikereiche:
The two second duration results in the test failing occasionally.  (KV timeout is 2.5 seconds).

babltiga:
It seems introduced with my test case, let me try to fix this one ...

babltiga:
Does it fail on data retrieval or on data cleanup? Sadly Jenkins is not publicly accessible for me to see the logs ...

mikereiche:
Sadly, the integration tests are not ran by jenkins as they need a cb server, and right now the runs in jenkings run without a cb server.

To run theses tests, it's necessary to modify the src/test/resources to point to a running cb server ( change the type=mocked to type=unmanaged and set the username/password and connection string to the cb server, then mvn integration-test)

All that it needs is the two second duration increased to 5 seconds or so.

mikereiche:
fixed along with  #1663

PR #1663, opened 2023-02-14
Bump couchbase sdk to 3_4_3.

### LINKED ISSUE #1661: bump couchbase java sdk to 3.4.3

mikereiche:
Requires some refactoring around @Stability.Internal APIs. Also fixed a test to get it to pass.

Closes #1661,#1662.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
+++ b/src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
@@ -15,6 +15,8 @@
  */
 package org.springframework.data.couchbase.core;
 
+import com.couchbase.client.core.api.query.CoreQueryContext;
+import com.couchbase.client.core.io.CollectionIdentifier;
 import reactor.core.publisher.Flux;
 
 import java.util.Optional;
@@ -28,7 +30,6 @@ import org.springframework.data.couchbase.core.support.PseudoArgs;
 import org.springframework.data.couchbase.core.support.TemplateUtils;
 import org.springframework.util.Assert;
 
-import com.couchbase.client.core.deps.com.fasterxml.jackson.databind.node.ObjectNode;
 import com.couchbase.client.java.ReactiveScope;
 import com.couchbase.client.java.json.JsonObject;
 import com.couchbase.client.java.query.QueryOptions;
@@ -100,11 +101,10 @@ public class ReactiveRemoveByQueryOperationSupport implements ReactiveRemoveByQu
 				} else {
 					TransactionQueryOptions opts = OptionsBuilder
 							.buildTransactionQueryOptions(buildQueryOptions(pArgs.getOptions()));
-					ObjectNode convertedOptions = com.couchbase.client.java.transactions.internal.OptionsUtil
-							.createTransactionOptions(pArgs.getScope() == null ? null : rs, statement, opts);
+					CoreQueryContext queryContext = CollectionIdentifier.DEFAULT_SCOPE.equals(rs.name()) ? null : CoreQueryContext.of(rs.bucketName(), rs.name());
 					return transactionContext.get().getCore()
-							.queryBlocking(statement, template.getBucketName(), pArgs.getScope(), convertedOptions, false)
-							.flatMapIterable(result -> result.rows).map(row -> {
+							.queryBlocking(statement, queryContext, opts.builder().build(), false)
+							.flatMapIterable(result -> result.collectRows()).map(row -> {
 								JsonObject json = JsonObject.fromJson(row.data());
 								return new RemoveResult(json.getString(TemplateUtils.SELECT_ID), json.getLong(TemplateUtils.SELECT_CAS),
 										Optional.empty());
--- a/src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
+++ b/src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.data.couchbase.core.query;
 
+import static com.couchbase.client.core.util.Validators.notNull;
 import static org.springframework.data.couchbase.core.query.Meta.MetaKey.RETRY_STRATEGY;
 import static org.springframework.data.couchbase.core.query.Meta.MetaKey.SCAN_CONSISTENCY;
 import static org.springframework.data.couchbase.core.query.Meta.MetaKey.TIMEOUT;
@@ -24,9 +25,13 @@ import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.time.Duration;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
+import com.couchbase.client.core.api.query.CoreQueryScanConsistency;
+import com.couchbase.client.core.classic.query.ClassicCoreQueryOps;
+import com.couchbase.client.core.error.InvalidArgumentException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.core.annotation.AnnotatedElementUtils;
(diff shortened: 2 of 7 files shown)


mikereiche (commented):
@dnault  @programmatix  for review.

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -67,21 +73,21 @@ public class OptionsBuilder {
 	static QueryOptions buildQueryOptions(Query query, QueryOptions options, QueryScanConsistency scanConsistency) {
 		options = options != null ? options : QueryOptions.queryOptions();
 		if (query.getParameters() != null) {
-			if (query.getParameters() instanceof JsonArray) {
+			if (query.getParameters() instanceof JsonArray && !((JsonArray) query.getParameters()).isEmpty()) {
 				options.parameters((JsonArray) query.getParameters());
-			} else {
+			} else if( query.getParameters() instanceof JsonObject && !((JsonObject)query.getParameters()).isEmpty()){
```
dnault, line 78: NIT: formatting

IntelliJ tip: Command-Option-Shift-L, then select "Only changes uncommitted to VCS"

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -370,10 +385,8 @@ static String toString(MutateInOptions o) {
 		return s.toString();
 	}
 
-	private static JsonObject getQueryOpts(QueryOptions.Built optsBuilt) {
-		JsonObject jo = JsonObject.create();
-		optsBuilt.injectParams(jo);
-		return jo;
+  public static JsonObject getQueryOpts(QueryOptions.Built optsBuilt) {
+    return JsonObject.fromJson(ClassicCoreQueryOps.convertOptions(optsBuilt).toString().getBytes());
```
dnault, line 389: getBytes(StandardCharsets.UTF_8)

src/test/java/org/springframework/data/couchbase/util/UnmanagedTestCluster.java
```diff
@@ -93,7 +94,7 @@ TestClusterConfig _start() throws Exception {
 		// no means to create a bucket on Capella
 		// have not created config() yet.
 		boolean usingCloud = seedHost.endsWith("cloud.couchbase.com");
-		bucketname = usingCloud ? "my_bucket" : UUID.randomUUID().toString();
+		bucketname = "my_bucket"; // usingCloud ? "my_bucket" : UUID.randomUUID().toString();
```
dnault, line 97: NIT: remove commented-out code?

src/test/java/org/springframework/data/couchbase/core/CouchbaseTemplateKeyValueIntegrationTests.java
```diff
@@ -94,8 +94,15 @@ void findByIdWithLock() {
...
 		} finally {
-			sleepSecs(2);
-			couchbaseTemplate.removeByQuery(User.class).withConsistency(QueryScanConsistency.REQUEST_PLUS).all();
+			for(int i=0; i< 10; i++) {
+				sleepSecs(2);
+				try {
+					couchbaseTemplate.removeByQuery(User.class).withConsistency(QueryScanConsistency.REQUEST_PLUS).all();
+					break;
+				} catch (Exception e) {
+					e.printStackTrace(); // gives IndexFailureException if the lock is still active
```
dnault, line 103: Does this hide other exceptions that aren't expected? (Sure, it logs the stack trace, but nobody's going to look at it if the test "passes".)

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -24,9 +24,15 @@
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.time.Duration;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 
+import com.couchbase.client.core.Core;
```
dnault, line 31: Unused import? (not just a NIT, because Core is potentially on the chopping block)

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -110,8 +116,17 @@ public static TransactionQueryOptions buildTransactionQueryOptions(QueryOptions
 			throw new IllegalArgumentException("QueryOptions.flexIndex is not supported in a transaction");
 		}
 
+		Object value = optsJson.get("args");
+		if(value instanceof JsonObject){
+			txOptions.parameters((JsonObject)value);
+		}else if(value instanceof JsonArray) {
+			txOptions.parameters((JsonArray) value);
+		}
```
dnault, line 124: else throw if not null?

dnault (changes requested):

src/main/java/org/springframework/data/couchbase/core/ReactiveRemoveByQueryOperationSupport.java
```diff
@@ -100,11 +101,10 @@ public Flux<RemoveResult> all() {
 				} else {
 					TransactionQueryOptions opts = OptionsBuilder
 							.buildTransactionQueryOptions(buildQueryOptions(pArgs.getOptions()));
-					ObjectNode convertedOptions = com.couchbase.client.java.transactions.internal.OptionsUtil
-							.createTransactionOptions(pArgs.getScope() == null ? null : rs, statement, opts);
+					CoreQueryContext queryContext = CollectionIdentifier.DEFAULT_SCOPE.equals(rs.name()) ? null : CoreQueryContext.of(rs.bucketName(), rs.name());
 					return transactionContext.get().getCore()
-							.queryBlocking(statement, template.getBucketName(), pArgs.getScope(), convertedOptions, false)
```
programmatix, line 106: I forgot that Spring was using these internals directly.  Let's get these methods annotated with the new annotation so breakage like this doesn't happen again.

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -110,8 +116,17 @@ public static TransactionQueryOptions buildTransactionQueryOptions(QueryOptions
...
+
 		for (Map.Entry<String, Object> entry : optsJson.toMap().entrySet()) {
-			txOptions.raw(entry.getKey(), entry.getValue());
+			if(!entry.getKey().equals("args")) {
+				txOptions.raw(entry.getKey(), entry.getValue());
```
programmatix, line 128: Continue to not feel good about this approach to clone a QueryOptions into a TransactionQueryOptions...  It won't handle things like the JsonSerializer, for example.

src/main/java/org/springframework/data/couchbase/core/query/OptionsBuilder.java
```diff
@@ -370,10 +385,8 @@ static String toString(MutateInOptions o) {
 		return s.toString();
 	}
 
-	private static JsonObject getQueryOpts(QueryOptions.Built optsBuilt) {
-		JsonObject jo = JsonObject.create();
-		optsBuilt.injectParams(jo);
-		return jo;
+  public static JsonObject getQueryOpts(QueryOptions.Built optsBuilt) {
```
programmatix, line 388: It doesn't feel great turning QueryOptions into JSON, and it also calls into the internal method ClassicCoreQueryOps.convertOptions,   Luckily, I don't think this getQueryOpts is necessary anymore.  It's called from two places:

1. buildQueryOptions.  Which only uses it for getScanConsistency.  But that's not needed anymore.  QueryOptions.Built now implements CoreQueryOptions so you can call scanConsistency() directly on it.
2. buildTransactionQueryOptions.  As mentioned above I already feel a bit queasy about this solution of converting a QueryOptions into JSON and then using that to set raw parameters on a TransactionQueryOptions.  And now with CoreQueryOptions  you don't need it - you can copy over the parameters one by one (now including the serializer).

programmatix (changes requested):

dnault (approved):

PR #1722, opened 2023-04-21
Add YearMonth converter.

### LINKED ISSUE #1720: Support for YearMonth Reading and Writer Converters
Hi, wanted to ask if this library can support the java.time.YearMonth class as a custom date converters.  This object is very useful for storing credit card expiration dates that only require year and month with no other time unit.    We have been using all of the out of box converters and this is the only one we have left that requires us to implement our own custom converter and think it would be useful for other users.    This is what we have and then override customConversions() in the AbstractCouchbaseConfiguration class.     Many thanks for considering it!

    @WritingConverter
    public enum YearMonthToStringConverter implements Converter<YearMonth, String> {

        INSTANCE;

        @Override
        public String convert(YearMonth source) {
            return Optional.ofNullable(source).map(YearMonth::toString).orElse(null);
        }
    }

    @ReadingConverter
    public enum StringToYearMonthConverter implements Converter<String, YearMonth> {

        INSTANCE;

        @Override
        public YearMonth convert(String source) {
            return Optional.ofNullable(source).map(YearMonth::parse).orElse(null);
        }
    }

mikereiche:
Closes #1720.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
+++ b/src/main/java/org/springframework/data/couchbase/core/convert/OtherConverters.java
@@ -22,6 +22,7 @@ import java.io.Writer;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
+import java.time.YearMonth;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


cabbonizio:
@mikereiche null checks needed at all like the other methods?   Concerned about potential NPEs

PR #1730, opened 2023-05-08
Automatically set readonly query option for read only queries

### LINKED ISSUE #1724: Automatically set QueryOptions readonly
When querying through (Reactive)CouchbaseTemplate, should query/count methods like `all()`, `one()`, `count()`, etc. automatically set the QueryOptions readonly to true?

> In addition to providing some security in that you are not accidentally modifying data, setting this flag to true
> also helps the client to more proactively retry and re-dispatch a query since then it can be sure it is idempotent.
> As a result, if your query is readonly then it is a good idea to set this flag.

micopiira:
Closes #1724


mikereiche:
This is going to change the exception thrown for timeouts from being an AmbiguousTimeoutException to UnambiguousTimeoutException. I'll merge it into main, but not merge it into 5.0.x.  I'll wait for 5.1.x

PR #1891, opened 2024-01-03
Callbacks refactoring

micopiira:
Aim of this PR is to bring the callbacks & events functionality in spring-data-couchbase
on par with other spring data projects, such as:
- spring-data-mongodb (https://docs.spring.io/spring-data/mongodb/docs/current/api/org/springframework/data/mongodb/core/mapping/event/package-summary.html)
- spring-data-r2dbc (https://docs.spring.io/spring-data/r2dbc/docs/current/api/org/springframework/data/r2dbc/mapping/event/package-summary.html)

**WARNING**: This PR contains a breaking changes:

Before this PR, the AfterConvertCallback was called before
save operations. This is no longer the case as
AfterConvertCallback is now only called after find operations.
Applications currently using AfterConvertCallbacks should
use BeforeSaveCallback instead. This change was done to be
consistent with other spring data projects.

### during Reactive save operations

- BeforeConvertEvent
- ReactiveBeforeConvertCallback
- BeforeSaveEvent
- ReactiveBeforeSaveCallback
- **save to db**
- AfterSaveEvent
- ReactiveAfterSaveCallback

### during Reactive fetch operations

- **fetch from db**
- AfterConvertEvent
- ReactiveAfterConvertCallback

### during blocking save operations

- BeforeConvertEvent
- BeforeConvertCallback
- BeforeSaveEvent
- BeforeSaveCallback
- **save to db**
- AfterSaveEvent
- AfterSaveCallback

### during blocking fetch operations

- **fetch from db**
- AfterConvertEvent
- AfterConvertCallback

TODO:
- Update reference documentation
- Bump major version?

Fixes https://github.com/spring-projects/spring-data-couchbase/issues/1854


mikereiche:
Thanks for offering this pull request. Breaking changes can only go in major versions.

If you make a PR with only "To be able to use spring-data-couchbase I would need an easy way to modify entities after fetching them from the database, as currently there is no easy way to do this are you interested in implementing new callbacks for modifying entities after find operations", then I will review that.  The callback can be named something like AfterDeserializationCallback so it does not conflict with AfterConvertCallback that currently serves another purpose. Please do not include other changes (such as updating comments or removing classes - that can go in separate PRs).

I appreciate the effort to unify the callback class names.  But if the class names are going to be changed, they should be changed to names that make sense.  For instance,  one could expect the calls

BeforeConvertCallback
-- convert --
AfterConvertCallback
-- save --

but that is not the case for mongodb. Instead they have

BeforeConvertCallback
-- converter --
BeforeSaveCallback
-- save --

Which is confusing because both callback are before the save. As far as I can tell, r2dbc was copied from mongodb resulting in the same problem.

If we're going to the trouble of renaming, why not   BeforeSerializationCallback, AfterSerializationCallback on write, and then on read BeforeDeserializationCallback and AfterDeserializationCallback on read()?

PR #1991, opened 2024-10-16
fix flaky test idHasMultiplePrefixesAndSuffixesSameOrder

ljmcr:
The flaky test idHasMultiplePrefixesAndSuffixesSameOrder was founded and detected using [NonDex](https://github.com/TestingResearchIllinois/NonDex) by running:
`mvn -pl drools-model/drools-model-codegen edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.springframework.data.couchbase.core.mapping.MappingCouchbaseConverterTests#idHasMultiplePrefixesAndSuffixesSameOrder`
In the test method, all `prefix` and `suffix` are added in the same order, which can cause `entity.id` to vary across different runs. I propose setting the `entity.id` to a fixed value to align with the test's intended behavior. Given that there is also a comment suggesting that the entire test may be disallowed, I would appreciate any feedback on how to best address this issue.

### DIFF
--- a/src/test/java/org/springframework/data/couchbase/core/mapping/MappingCouchbaseConverterTests.java
+++ b/src/test/java/org/springframework/data/couchbase/core/mapping/MappingCouchbaseConverterTests.java
@@ -1184,10 +1184,12 @@ public class MappingCouchbaseConverterTests {
 			@IdSuffix(order = 1) public String suffix = "222";
 		}
 		Entity entity = new Entity();
+		entity.id = entity.prefix + '.' + entity.prefix1 + '.' + entity.someId + '.' + entity.suffix + '.' + entity.suffix1;
 		CouchbaseDocument converted = new CouchbaseDocument();
 		converter.write(entity, converted);
 		assertThat(converted.getId()).isEqualTo(entity.id);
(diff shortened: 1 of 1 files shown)


src/test/java/org/springframework/data/couchbase/core/mapping/MappingCouchbaseConverterTests.java
```diff
@@ -1184,10 +1184,12 @@ class Entity {
 			@IdSuffix(order = 1) public String suffix = "222";
 		}
 		Entity entity = new Entity();
+		entity.id = entity.prefix + '.' + entity.prefix1 + '.' + entity.someId + '.' + entity.suffix + '.' + entity.suffix1;
```
mikereiche, line 1187: The test is for the construction of converted.getId() by the converter.  When entity.id is already it us used instead of constructing it.

mikereiche (commented):
> The flaky test idHasMultiplePrefixesAndSuffixesSameOrder 

How is it flaky? It always passes with the correct result.

> prefix and suffix are added in the same order, which can cause entity.id to vary across different runs

putting them in the same order is what makes entity.id always "333.abc.222"

PR #2042, opened 2025-04-01
Enhance CouchbaseDocument and CouchbaseList to handle List implementa…

teetangh:
# Fix for Collections.SingletonList handling in Spring Data Couchbase

This PR addresses an issue where `Collections.SingletonList` and other List implementations cannot be stored directly in Couchbase documents after upgrading to Spring Boot 3.4.3 with Spring Data Couchbase 5.4.3.

## Details

The error manifests as:
```
java.lang.IllegalArgumentException: Attribute of type java.util.Collections.SingletonList cannot be stored and must be converted.
```

The fix adds special handling for List implementations in the `verifyValueType` methods of both `CouchbaseDocument` and `CouchbaseList` classes, allowing them to be processed without requiring explicit conversion.

## Changes

- Modified `CouchbaseDocument.verifyValueType()` to accept `List` implementations
- Modified `CouchbaseList.verifyValueType()` to accept `List` implementations
- Updated error messages to provide clarity about which component is generating the error

This change ensures backward compatibility with code that worked in Spring Boot 3.2.4 while maintaining proper functionality in Spring Boot 3.4.3.

### DIFF
--- a/src/main/java/org/springframework/data/couchbase/core/mapping/CouchbaseDocument.java
+++ b/src/main/java/org/springframework/data/couchbase/core/mapping/CouchbaseDocument.java
@@ -20,6 +20,7 @@ import com.couchbase.client.java.json.JsonObject;
 
 import java.util.TreeMap;
 import java.util.Map;
+import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


mikereiche (commented):
Needs a test case.

